PR #3, opened 2021-09-08
issue/2 Check all previous locks

### LINKED ISSUE #2: Locking should not lock completed items
For example: sequential locking requires the completion of the previous item. If the current item is already complete, it should not be locked.

This is particularly awkward if a user rests completion on a component in a previous topic. It then causes subsequent menu items to relock even when previously completed.

oliverfoster:
fixes #2 

### Fixed
* Locking is aware of previous locking state, if the previous item _isLocked then this item will be locked

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -718,25 +718,31 @@ export default class AdaptModel extends LockingModel {
 
   setSequentialLocking() {
     const children = this.getAvailableChildModels();
-    const firstChild = children.shift();
-    children.reduce((previousChild, child) => {
-      const isLockedByPreviousChild = (!previousChild.get('_isComplete') && !previousChild.get('_isOptional'));
-      return child.set('_isLocked', isLockedByPreviousChild && !child.get('_isComplete'));
-    }, firstChild);
+    // Start from second child
+    children.slice(1).forEach((child, index) => {
+      const previousChild = children[index];
+      // If previous was locked, all subsequent will be locked.
+      const isLockedByPreviousChild = previousChild.get('_isLocked') ||
+        (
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

danielghost (commented):
This code works as explained but I think this may cause other UX issues as explained here:
https://github.com/adaptlearning/adapt-contrib-core/issues/2#issuecomment-916019260

oliverfoster:
I think we have agreed that all previous locks should be checked rather than just the last previous lock. PR needs modifying.

oliverfoster:
Updated according to https://github.com/adaptlearning/adapt-contrib-core/issues/2#issuecomment-916121035

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

js/models/adaptModel.js
```diff
@@ -749,8 +755,13 @@ export default class AdaptModel extends LockingModel {
     if (!lockedBy) return false;
     return lockedBy.some(id => {
       try {
-        const model = Adapt.findById(id);
-        return !model.get('_isComplete') && !model.get('_isOptional') && model.get('_isAvailable');
+        const anotherModel = Adapt.findById(id);
+        return anotherModel.get('_isLocked') ||
```
oliverfoster, line 759: Is it worth putting an availability check in here?

oliverfoster: Done

joe-allen-89 (approved):
👀

PR #5, opened 2021-09-15
issue/4 Added Adapt.a11y.ariaLevel

### LINKED ISSUE #4: Aria level calculation should be accessible to react
Currently the helper [`a11y_attrs_heading` ](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/helpers.js#L341) produces a string of attributes which isn't compatible with react.

oliverfoster:
fixes #4 

### Changed
* Moved aria level calculation from `a11y_attrs_heading` helper to `Adapt.a11y.ariaLevel`
* Fixed some linting issues

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -154,31 +154,32 @@ class A11y extends Backbone.Controller {
 
   /**
    * Calculate the aria level for a heading
-   * @param {string|number} [levelOrType] Specify a level or type group.
+   * @param {string|number} [defaultLevelOrType] Specify a default level or type group.
+   * @param {string|number} [overrideLevelOrType] Override with a level or type group from the designer.
    * @returns {number}
    */
-  ariaLevel(levelOrType) {
+  ariaLevel(defaultLevelOrType = 1, overrideLevelOrType) {
     // get the global configuration from config.json
     const cfg = Adapt.config.get('_accessibility');
-    // default level to use if nothing overrides it
-    let level = 1;
 
     // first check to see if the Handlebars context has an override
-    if (this._ariaLevel) {
-      levelOrType = this._ariaLevel;
+    if (overrideLevelOrType) {
+      defaultLevelOrType = overrideLevelOrType;
     }
 
-    if (isNaN(levelOrType) === false) {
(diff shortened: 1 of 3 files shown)


cahirodoherty-learningpool (approved):
👀

js/a11y.js
```diff
@@ -151,6 +152,35 @@ class A11y extends Backbone.Controller {
...
+    let level = 1;
+
+    // first check to see if the Handlebars context has an override
+    if (this._ariaLevel) {
+      levelOrType = this._ariaLevel;
+    }
+
+    if (isNaN(levelOrType) === false) {
+      // if a number is passed just use this
+      level = levelOrType;
```
eleanor-heath, line 173: return early here

js/a11y.js
```diff
@@ -151,6 +152,35 @@ class A11y extends Backbone.Controller {
...
+    } else if (_.isString(levelOrType)) {
```
eleanor-heath, line 174: this can be an `if` instead of `else if`

js/a11y.js
```diff
@@ -151,6 +152,35 @@ class A11y extends Backbone.Controller {
...
+      // if a string is passed check if it is defined in global configuration
+      cfg._ariaLevels = cfg._ariaLevels || defaultAriaLevels;
+      if (cfg._ariaLevels?.['_' + levelOrType] !== undefined) {
```
eleanor-heath, line 177: return early here?

eleanor-heath (approved):
left a few suggestions around returning early but otherwise looks good.

js/helpers.js
```diff
@@ -1,16 +1,5 @@
...
-  '_menuGroup': 2,
-  '_menuItem': 2,
-  '_page': 1,
-  '_article': 2,
-  '_block': 3,
-  '_component': 4,
-  '_componentItem': 5,
-  '_notify': 1
-};
+import a11y from './a11y';
```
eleanor-heath, line 2: should a11y be capitalised to A11y?

oliverfoster: no, it's an instance at this point, not a class.

chris-steele (approved):

PR #7, opened 2021-09-16
issue/6 Distinction between readable and focusable for findForward

### LINKED ISSUE #6: Aria grouping roles should not receive focus
With new mcq and gmcq, the options are grouped with a parent having `[role=group]` or `[role=radiogroup]` and `[aria-labelledby]`. 

Adapt's focus functions will automatically focus on `[aria-labelledby]` elements when ask to `focusFirst` or `focusNext` due to this [configuration option](https://github.com/adaptlearning/adapt-contrib-core/blob/0dacbfbe0215c452c9b2bc4a96db44c38f37a6b9/js/a11y.js#L42).

As the container receives focus, with a screen reader, the entire mcq or gmcq will read once. Then upon keyboard navigation the mcq will read again.

It should be constrained to exclude something like `[aria-labelledby]:not([role=radiogroup],[role=group],[role=tablist],[role=list],[role=tree],[role=treegrid],[role=table],[role=grid],[role=menu],[role=rowgroup])`, explicitly preventing Adapt from focusing on grouping roles.

Reference https://www.w3.org/TR/wai-aria-1.1/#role_definitions

oliverfoster:
fixes #6 

### Added
* `isFocusable, findFocuable, findFirstFocuable` as a set of functions to complement `isReadable, findReadable, findFirstReadable` and `isTabbable, findTabbable, findFirstTabbable`

### Fixed
* `focusNext` and `focusFirst` now use `findFirstFocusable` instead of `findFirstReadable` to excludes grouping elements allowing content to be focused rather than the container

### Changed
* Return `null` instead of `undefined` for clarity

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -42,6 +42,10 @@ class A11y extends Backbone.Controller {
       _focusableElements: 'a,button,input,select,textarea,[tabindex],label',
       _readableElements: '[role=heading],[aria-label],[aria-labelledby],[alt]',
       /**
+       * Prevent adapt from focusing forward onto these elements.
+       */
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

joe-allen-89 (approved):

PR #15, opened 2021-09-29
issue/2967 Reworked scrollingContainer

oliverfoster:
fixes https://github.com/adaptlearning/adapt_framework/issues/2967

### Fixed
* Replaced jquery based fix with browser based fix to enhance the embeddedness of the solution
  * Replaced `window.scrollY`, `window.scrollX`, `window.pageYOffset`, `window.pageXOffset`, `html.scrollHeight`, `html.scrollWidth` and `window.scrollTo` so that they have impact on the `#app` container and use its values rather than those from the `html` tag
* Added `MouseEvent` fix to ensure `event.pageX` and `event.pageY` are relative to the scroll of the `#app` container rather than the `html` tag
  * Replaced `MouseEvent.prototype.pageX` and `MouseEvent.prototype.pageY` so that they take the approprate values from the `#app` container scroll position rather than the `html` tag scroll position

### Changed
* Linting issues

### Testing
1. Clone the framework and set core to the correct branch:
```sh
git clone https://github.com/adaptlearning/adapt_framework 2967
cd 2967
npm install && adapt devinstall
cd src/core
git checkout issue/2967
cd ../../
```

2. Enable the scrolling container by changing `src/course/config.json` `_scrollingContainer._isEnabled` to `true`.
3. Add volume controls to the media component by adding to `src/course/en/components.json` component `c-40`, `"_showVolumeControl": true,`

3. Build the project:
```sh
grunt dev
```

PR #16, opened 2021-09-29
issue/2967-b Scrolling container rewrite

oliverfoster:
fixes https://github.com/adaptlearning/adapt_framework/issues/2967

alternative to https://github.com/adaptlearning/adapt-contrib-core/pull/15

### Differences
* The scrolling container will be enabled by default
* Instead of scrolling using the viewport to scroll, Adapt would use the `body` tag to scroll rather than `#app` tag as before 
* The navigation bar can remain fixed position rather than being relative, as it was with the old implementation
* The content can reach from the top of the screen under the navigation bar, to the bottom screen, rather than from under the navigation bar to the bottom of the screen, which brings the behaviour inline in both modes
* The navigation bar width must be offset for the `body` element's scrollbar, this is as unlike the viewport scrollbar the body scrollbar does not automatically constrain the fixed position navigation bar width by those extra 18px
* The scrollbar width is calculated automatically and is stored in a css variable so that the browser can apply the styling to the navigation bar automatically (IE11 has a fallback style for 18px, which never changes, Safari has floating scrollbars and so is always calculated as 0px)

### Impact
* With both the scrolling container enabled and disabled, Adapt can be styled identically, which it could not be before [see here](https://github.com/adaptlearning/adapt_framework/issues/2710)
* Adapt will work in ios iframes by default
* [This new ios bug](https://github.com/adaptlearning/adapt-contrib-trickle/issues/132) (presenting in trickle) will be fixed, which occurs as there is a 44px zone at the bottom of mobile safari that prevents the trickle button from being clicked ([referenced here](https://www.eventbrite.com/engineering/mobile-safari-why/)) but the click instead activates the bottom browser menu. With the scrolling container enabled, this pr forces the browser navigation bars to be always present

### Fixed
* Replaced jquery based fix with browser based fix to enhance the embeddedness of the solution
  * Replaced `window.scrollY`, `window.scrollX`, `window.pageYOffset`, `window.pageXOffset`, `html.scrollHeight`, `html.scrollWidth` and `window.scrollTo` so that they have impact on the `body` container and use its values rather than those from the `html` tag
* Added `MouseEvent` fix to ensure `event.pageX` and `event.pageY` are relative to the scroll of the `body` container rather than the `html` tag
  * Replaced `MouseEvent.prototype.pageX` and `MouseEvent.prototype.pageY` so that they take the approprate values from the `body` container scroll position rather than the `html` tag scroll position  

### Changed
* Linting issues

### Testing
1. Clone the framework and set core to the correct branch:
```sh
git clone https://github.com/adaptlearning/adapt_framework 2967-b
cd 2967-b
npm install && adapt devinstall
cd src/core
git checkout issue/2967-b
cd ../../
```

2. Enable the scrolling container by changing `src/course/config.json` `_scrollingContainer._isEnabled` to `true`.
3. Add volume controls to the media component by adding to `src/course/en/components.json` component `c-40`, `"_showVolumeControl": true,`

3. Build the project:
```sh
grunt dev
```

### ios iframe test
[iframe.zip](https://github.com/adaptlearning/adapt-contrib-core/files/7252103/iframe.zip)

Place the `iframe.html` file contained in the above zip, alongside `index.html` in the build folder. Open the `iframe.html` on an iphone/ipad.

Or upload to an lms which embeds the course in an iframe

### DIFF
--- a/less/core/scrolling.less
+++ b/less/core/scrolling.less
@@ -8,22 +8,19 @@ html.adapt-scrolling {
   // Allow overflow scroll top for refresh on android
   overflow-y: auto;
 
-  body, #app {
-    .restrain-height;
-  }
-
+  // Offset the navigation bar width the the body scrollbar
   .nav {
-    position: relative;
+    // IE11 is consistant at 18px
+    width: calc(100% - 18px);
+    // Modern browsers change their scroll bar width on zoom
+    width: calc(100% - var(--adapt-scrollbar-width));
   }
 
-  #app {
+  body {
+    .restrain-height;
     position: relative;
-    height: calc(~'100% - @{navigation-height}');
     overflow-y: scroll;
+    // Allow scrolling of the body in an iframe on ios
     -webkit-overflow-scrolling: touch;
-  }
-
-  #wrapper {
-    padding: 0;
   }
 }
--- a/less/core/scrolling.less
+++ b/less/core/scrolling.less
@@ -1,12 +1,7 @@
-.restrain-height() {
-  height: 100%;
-  overflow: hidden;
-}
-
 html.adapt-scrolling {
-  .restrain-height;
+  height: 100%;
   // Allow overflow scroll top for refresh on android
-  overflow-y: auto;
+  overflow: auto;
 
   // Offset the navigation bar width the the body scrollbar
   .nav {
@@ -17,7 +12,8 @@ html.adapt-scrolling {
   }
 
   body {
-    .restrain-height;
+    height: 100%;
+    overflow: hidden;
     position: relative;
     overflow-y: scroll;
     // Allow scrolling of the body in an iframe on ios
--- a/less/core/scrolling.less
+++ b/less/core/scrolling.less
@@ -5,6 +5,8 @@
 
 html.adapt-scrolling {
   .restrain-height;
+  // Allow overflow scroll top for refresh on android
+  overflow-y: auto;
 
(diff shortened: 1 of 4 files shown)


less/core/scrolling.less
```diff
@@ -1,27 +1,22 @@
...
-  }
+  height: 100%;
+  // Allow overflow scroll top for refresh on android
+  overflow: auto;
 
+  // Offset the navigation bar width the the body scrollbar
   .nav {
-    position: relative;
+    // IE11 is consistant at 18px
+    width: calc(100% - 18px);
```
eleanor-heath, line 9: should 18px not be converted into rem?

eleanor-heath (approved):

less/core/scrolling.less
oliverfoster: No, the browser scrollbar in IE11 doesn't change with the body font size.

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

ethan-lp (approved):

oliverfoster:
One +1 from kineo would be good please? @chris-steele @danielghost @kirsty-hames @guywillis

kirsty-hames (approved):
Nice scrollbar width fix @oliverfoster and can confirm [#2967](adaptlearning/adapt_framework#2967) fixed.

PR #21, opened 2021-10-14
issue/2973

cahirodoherty-learningpool:
Check the newLangugae is not equal to the current language before triggering app:languageChanged

[2973](https://github.com/adaptlearning/adapt-contrib-core/issues/188)

### DIFF
--- a/js/data.js
+++ b/js/data.js
@@ -82,20 +82,16 @@ class Data extends AdaptCollection {
     this.loadCourseData(language);
   }
 
-  async loadCourseData(newLanguage) {
-
+  async loadCourseData(newLanguage = Adapt.config.get('_activeLanguage')) {
     // All code that needs to run before adapt starts should go here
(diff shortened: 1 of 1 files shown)


js/data.js
```diff
@@ -166,7 +166,7 @@ class Data extends AdaptCollection {
   }
 
   async triggerDataReady(newLanguage) {
-    if (newLanguage) {
+    if (newLanguage && newLanguage !== Adapt.config.get('_activeLanguage')) {
```
oliverfoster, line 169:
```suggestion
    const hasLanguageChanged = (newLanguage !== Adapt.config.get('_activeLanguage'));
    if (newLanguage && hasLanguageChanged) {
```

cahirodoherty-learningpool: Thanks Oliver. I think I will deleting this draft PR though after all. The language comparison at this point is no good to us as they will always be equal

PR #44, opened 2021-11-16
issues/10

### LINKED ISSUE #10: LMS launch in same window breaks when lang param is set
There seems to be (intermittent) buggy behaviour in the case of launching Adapt content in the current window when the lang param is set. The content seems to launch and close instantly.

On debugging I have found that what is actually happening is that the onLanguageChanged that is triggered is calling a second routers.handleRoute. The first pass of this function will have a false value when checking this.model.get('_canNavigate') (https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/router.js#L92), but the second pass is true. This is because inside the if (first pass) the value is set to false (https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/router.js#L94).

Overall this results in a `Backbone.history.history.back()` call (https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/router.js#L116), forcing the window to return to the LMS launch screen

FW5.15.5

Possibly related to https://github.com/adaptlearning/adapt_framework/issues/2973

cahirodoherty-learningpool:
Prevent triggerDataReady deferred navigation call if the router model doesn't allow it

fixes #10

### DIFF
--- a/js/data.js
+++ b/js/data.js
@@ -170,6 +170,9 @@ class Data extends AdaptCollection {
       Adapt.trigger('app:languageChanged', newLanguage);
       _.defer(() => {
         Adapt.startController.loadCourseData();
+
+        if (!Adapt.router.model.get('_canNavigate')) return;
+
         const hash = Adapt.startController.isEnabled() ? Adapt.startController.getStartHash(false) : '#/';
         Adapt.router.navigate(hash, { trigger: true, replace: true });
       });


eleanor-heath (approved):

ethan-lp (approved):

js/data.js
```diff
@@ -170,6 +170,9 @@ class Data extends AdaptCollection {
       Adapt.trigger('app:languageChanged', newLanguage);
       _.defer(() => {
         Adapt.startController.loadCourseData();
+
+        if (!Adapt.router.model.get('_canNavigate')) return;
```
oliverfoster, line 174: This is definitely the right place... I'd like to stop that double call to `routeId` if we can.

The double call seems to be happening because of [this defer](https://github.com/adaptlearning/adapt-contrib-core/blob/dc9e3c22acd67120d0cb7f7fd844e6df03a16f2e/js/data.js#L171) (which was inherited from [this ancient commit](https://github.com/adaptlearning/adapt-contrib-core/commit/50990850e561f12042446bcc5a4cec7deda3ec85#diff-32cdce28ac549728f6a28999f6d874aa4f29e6da1406c28c3134cd8d421a866aR56) through [here](https://github.com/adaptlearning/adapt-contrib-core/commit/36317b1c9626cd88e6a9dae12fc28694b480f5e6)) and the gap between [`_canNavigate: false`](https://github.com/adaptlearning/adapt-contrib-core/blob/dc9e3c22acd67120d0cb7f7fd844e6df03a16f2e/js/router.js#L94) and waiting for the content object to be ready as [`_canNavigate: true`](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/router.js#L189).

Effectively, it seems that because of the defer, it tries to route to `#/` or the start controller hash, almost immediately after the data is loaded and halfway through the content object readiness. I don't see any need for that defer now. If it's removed we can switch the router to the default hash before it has been through the first routing cycle.

Can you try this one https://github.com/adaptlearning/adapt-contrib-core/pull/45?

cahirodoherty-learningpool: Unfortunately, these changes lose the intention of the original work, i.e. force routing on language change (and ideally on `lang` param passing). Yours seems like the right approach, but we will need to catch that behaviour in some other part of the code. I'll continue to investigate

oliverfoster: I'm not sure I've understood. From my experimentation, it just does this in preparation for the first call to routeId rather than on a second. Do you want me to send over my test course? Shall we have a call about it perhaps? (I've got a cold so I'm a bit lost in translation.)

cahirodoherty-learningpool: My mistake! I hadn't the start controller configured for the other languages I was testing with, which is why I was seeing that behaviour. I will close this branch now that your work has been merged.
Thanks again

PR #48, opened 2021-11-24
issue/22 Throw error rendering past locked content

### LINKED ISSUE #22: Navigating to locked models
`navigateToElement` in the router, allows scrolling to locked models. Previously this may have been prevented by https://github.com/adaptlearning/adapt-contrib-trickle, but this doesn't seem to be the case with the most recent version. Should core handle this rather than trickle?

### Your environment
* FW 5.17.1
* adapt-contrib-trickle 5.2.2

### Steps to replicate
* Fail an assessment and exit the course from that page
* Use the bookmark dialog to return to the previous location

oliverfoster:
fixes #22 
part fix for https://github.com/adaptlearning/adapt-contrib-bookmarking/issues/46
part fix for https://github.com/adaptlearning/adapt-contrib-trickle/issues/128

### Fixed
* Throw error when attempting to render past locked content

### DIFF
--- a/js/views/contentObjectView.js
+++ b/js/views/contentObjectView.js
@@ -111,6 +111,8 @@ export default class ContentObjectView extends AdaptView {
     }
     // Force all models up until the id to render
     models = models.slice(0, index + 1);
+    const isLocked = models.some(model => model.get('_isLocked'));
+    if (isLocked) throw new Error(`Cannot renderTo ${id} as it is preceded by locked content`);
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

lc-alexanderbenesch (approved):
👍

danielghost (commented):
Code looks good, but just referencing the following issue as this can make it look like this can be bypassed:
https://github.com/adaptlearning/adapt-contrib-branching/issues/17

Issue #55, opened 2022-01-10
Assessment not submitting score properly


oliverfoster:
Should the default “adapt create course” install submit a score on assessment completion to the LMS? This is v5.18.1 with the latest plugins and config._spoor._shouldSubmitScore = true.

I’m getting no score sent to SCORM cloud with this setup.

oliverfoster:
https://github.com/adaptlearning/adapt-contrib-spoor/commit/f3315d658b7c6f09b88bb9ad760cb6e5eb86728e
https://github.com/adaptlearning/adapt-contrib-core/pull/30

Prs moved the behaviour but not the `_shouldSubmitScore` property.

deltanetdan:
Tracking still needs _requireAssessmentCompleted = true on [this line ](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/tracking.js#L21)before it will send the score.

A user might want to setup the course to track course completion only and still send the score to the LMS.

PR #57, opened 2022-01-12
Allow score to be set if _requireAssessmentCompleted is false

deltanetdan:
Move the check for _requireAssessmentCompleted to onAssessmentComplete. Fix for #55

### DIFF
--- a/js/tracking.js
+++ b/js/tracking.js
@@ -17,13 +17,10 @@ class Tracking extends Backbone.Controller {
   }
 
   setupEventListeners() {
-    // Check if completion requires passing an assessment.
-    if (this._config._requireAssessmentCompleted) {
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

eleanor-heath (approved):

deltanetdan:
Actually I think _shouldSubmitScore needs adding to config.model.schema as well

oliverfoster:
I'm just waiting for a spec for the new schema from @taylortom before we go editing them. I guess it just follows json schema for the most part and should be a copy/paste from any other boolean property.

oliverfoster:
schema can be sorted in another pr

Issue #69, opened 2022-02-24
Component titles are not being read out by screenreaders on IOs Devices


eleanor-heath:
**Steps to replicate**
1. Create a course with a text component
2. Navigate to the text component title using a screen reader on an iOS device (Im using an iPhone 7)

**Expected Results**
The screenreader should read out the the component titles

**Actual Results**
The component titles are not being read out

**Additional Notes**
I am on framework 15.8.4

oliverfoster:
I'll try and give this a test this week, it seems odd that they're not working correctly. It's tested quite thoroughly.

Which device and ios version?

eleanor-heath:
@oliverfoster I tested on my iPhone 7 iOS 15.3.1. The issue was raised by the client for iPad/ioS v15.x/Safari/Voiceover screen reader.

oliverfoster:
Can't replicate:
https://user-images.githubusercontent.com/7974663/157737494-ade6a134-0283-4f53-b822-6f0c2a96c681.mp4

oliverfoster:
closed all tested fine

PR #70, opened 2022-02-24
ADAPT-9725 - ensure component titles can be detected by the voiceover screenreader on ios devices

eleanor-heath:
https://github.com/adaptlearning/adapt-contrib-core/issues/69
https://github.com/adaptlearning/adapt-contrib-core/issues/75

### DIFF
--- a/templates/header.jsx
+++ b/templates/header.jsx
@@ -39,6 +39,7 @@ export default function Header(props) {
   if (!isSet) return null;
   const _globals = Adapt.course.get('_globals');
   const ariaRegion = _globals?._components?.[`_${_component}`]?.ariaRegion;
+  const isIOSDevice = /(iPad|iPhone|iPod)/g.test(navigator.userAgent);
   return (
     <div id={`${_id}-header`} className={prefixClasses(classNamePrefixes, ['__header'])}>
       <div className={prefixClasses(classNamePrefixes, ['__header-inner'])}>
@@ -49,7 +50,7 @@ export default function Header(props) {
           <div className="js-heading" ref={jsxHeading}></div>
           }
 
-          <div className={prefixClasses(classNamePrefixes, ['__title-inner'])} aria-hidden={!_disableAccessibilityState}>
+          <div className={prefixClasses(classNamePrefixes, ['__title-inner'])} aria-hidden={isIOSDevice ? false : !_disableAccessibilityState}>
             {html(compile(displayTitle, props))}
           </div>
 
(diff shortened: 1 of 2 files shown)


lc-alexanderbenesch (approved):
👍
tested on iPhone 8, 15.3.1

templates/header.jsx
```diff
@@ -49,7 +50,7 @@ export default function Header(props) {
           <div className="js-heading" ref={jsxHeading}></div>
           }
 
-          <div className={prefixClasses(classNamePrefixes, ['__title-inner'])} aria-hidden={!_disableAccessibilityState}>
+          <div className={prefixClasses(classNamePrefixes, ['__title-inner'])} aria-hidden={isIOSDevice ? false : !_disableAccessibilityState}>
```
cahirodoherty-learningpool, line 53: Hi Eleanor,
Adapt.device already has a function to perform this check. It may be best to use this in case it ever gets improved in the future

cahirodoherty-learningpool: `<div className={prefixClasses(classNamePrefixes, ['__title-inner'])} aria-hidden={Adapt.device.isAppleDevice() ? false : !_disableAccessibilityState}>`

cahirodoherty-learningpool (changes requested):

templates/header.jsx
eleanor-heath: The only thing about using `Adapt.device.isAppleDevice()` is that Ive noticed before that it does not apply to iPad devices and will need to add an additional check to isAppleDevice, such as `navigator.userAgent.match(/Mac/) && navigator?.maxTouchPoints > 2;` to account for iPadOS. I will create another PR for this in the core repo now.

eleanor-heath: I have opened another issue to update `isAppleDevice` here https://github.com/adaptlearning/adapt-contrib-core/issues/75 and have included that fix in this pull request.

cahirodoherty-learningpool (approved):
👀

oliverfoster:
I need to replicate this bug on ios, I don't understand why it's not working as is at the moment.

oliverfoster:
Screen capture here shows that voiceover is reading the vanilla course titles just fine https://github.com/adaptlearning/adapt-contrib-core/issues/69#issuecomment-1064401749

oliverfoster:
Screen capture here shows that voiceover is reading the vanilla course titles just fine https://github.com/adaptlearning/adapt-contrib-core/issues/69#issuecomment-1064401749

oliverfoster:
tested all fine

PR #73, opened 2022-03-01
issue/62 Removed page:scrollTo from notifyPopupView removal

### LINKED ISSUE #62: notify close on `page:scrollTo`
Can the listener to close notify on `page:scrollTo` be removed?
https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/views/notifyView.js#L38

Was there a specific reason why it was added for this commit?
https://github.com/adaptlearning/adapt-contrib-core/blob/9e3deaf56fa7d57829caec9f5712a835825c0f14/js/views/notifyView.js#L38

This seems unnecessary as it shouldn't be possible to trigger this event with normal user behaviour, whilst a notification is displayed. If using notify to display error messages, i.e. LMS, communication errors etc., these can be forced closed when the event is fired on trickle autoscroll.

oliverfoster:
fixes #62 

### Removed
* Unusual event listener cause erroneous remove on notify

### DIFF
--- a/js/views/notifyPopupView.js
+++ b/js/views/notifyPopupView.js
@@ -36,7 +36,7 @@ export default class NotifyPopupView extends Backbone.View {
 
   setupEventListeners() {
     this.listenTo(Adapt, {
-      'remove page:scrollTo': this.closeNotify,
+      remove: this.closeNotify,
       'notify:resize': this.resetNotifySize,
       'notify:cancel': this.cancelNotify,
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

lc-alexanderbenesch (approved):
👍

joe-allen-89 (approved):

PR #94, opened 2022-04-13
issue/91 Use menu aria-level for course

### LINKED ISSUE #91: course aria-level
Currently the `js-heading-inner` will use `aria-level="course"` for a main menu, as the [config.json](https://github.com/adaptlearning/adapt_framework/blob/master/src/course/config.json#L14-L24) doesn't include a `"_course": 1` attribute. Should the `_course` type be added, or should the `a11y_attrs_heading` handlbars helper be modified so the `"_menu": 1` is applied to all menus regardless of the model type? I assume currently the `_menu` type will only be used for sub-menus?

oliverfoster:
fixes #91 

### Fixed
* Use the menu aria-level for the course menu


eleanor-heath (approved):

kirsty-hames (approved):
+1

PR #96, opened 2022-04-20
issue/95 Submit display none fix

### LINKED ISSUE #95: Focus goes to body if submit button is display: none on submit
Sometimes a theme will require the buttons to disappear once the submit button is clicked. This often means that the focus will drop back to the body element, which isn't ideal as it breaks the flow of the screen reader context.

oliverfoster:
fixes https://github.com/adaptlearning/adapt_framework/issues/3190
fixes #95 

The focus now moves forward to the marking aria label if the button goes missing.

### Fixed
* Moved the marking aria label to after the submit button to make it readable [3190](https://github.com/adaptlearning/adapt_framework/issues/3190)
* Blur handler now checks for `display: none` and `visibility: hidden` as well as the `[disabled]` attribute

### DIFF
--- a/js/a11y/browserFocus.js
+++ b/js/a11y/browserFocus.js
@@ -47,7 +47,7 @@ export default class BrowserFocus extends Backbone.Controller {
     }
     // Check if element losing focus is losing focus
     // due to the addition of a disabled class
-    if (!$element.is('[disabled]')) {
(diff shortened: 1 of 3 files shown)


StuartNicholls (approved):

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

PR #108, opened 2022-05-09
issue/3295 New list structure in drawer

joe-replin:
#3295 Making drawer__holder the list with a direct child, either a 'listitem' or 'group' depending on the current view

### DIFF
--- a/templates/drawer.hbs
+++ b/templates/drawer.hbs
@@ -2,7 +2,7 @@
 
   <span id="drawer-heading" class="aria-label">{{_globals._accessibility._ariaLabels.drawer}}</span>
 
-  <div class="drawer__holder js-drawer-holder"></div>
+  <div class="drawer__holder js-drawer-holder" role="list"></div>
 
   <div class="drawer__toolbar u-clearfix">
 


oliverfoster:
@joe-replin have you screen reader tested?

templates/drawer.hbs
```diff
@@ -2,7 +2,7 @@
 
   <span id="drawer-heading" class="aria-label">{{_globals._accessibility._ariaLabels.drawer}}</span>
 
-  <div class="drawer__holder js-drawer-holder"></div>
+  <div class="drawer__holder js-drawer-holder" role="list"></div>
```
oliverfoster, line 5: The roll here needs to be selectively applied as it works if you have search and resources installed, but not with just resources or in plp where this list reads with one item.

oliverfoster: if used drawerView.js, function renderItems() {}
```js
  renderItems() {
    Adapt.trigger('drawer:empty');
    this.emptyDrawer();
    const isList = (this.collection.length > 1);
    if (isList) this.$('.js-drawer-holder').attr('role', 'list');
    else this.$('.js-drawer-holder').removeAttr('role');
    this.collection.forEach(model => new DrawerItemView({ model }));
  }
```

And hideDrawer() removes the role=list in both animation and non-animation modes:
```js
hideDrawer($toElement) {
    if (!this._isVisible) return;

    const direction = {};
    a11y.popupClosed($toElement);
    this._isVisible = false;
    a11y.scrollEnable('body');

    if (this.disableAnimation) {

      direction[this.drawerDir] = -this.$el.width();
      this.$el
        .css(direction)
        .addClass('u-display-none')
        .attr('aria-hidden', 'true');

      $('.js-shadow').addClass('u-display-none');
      this.$('.js-drawer-holder').removeAttr('role');

      Adapt.trigger('drawer:closed');

    } else {
      const easing = Adapt.config.get('_drawer')?._hideEasing || 'easeInQuart';

      direction[this.drawerDir] = -this.$el.width();
      this.$el.velocity(direction, this.drawerDuration, easing, () => {
        this.$el
          .addClass('u-display-none')
          .attr('aria-hidden', 'true');
        this.$('.js-drawer-holder').removeAttr('role');
        Adapt.trigger('drawer:closed');
      });

      $('.js-shadow').velocity({ opacity: 0 }, {
        duration: this.drawerDuration,
        complete() {
          $('.js-shadow').addClass('u-display-none');
        }
      });

    }

    this._isCustomViewVisible = false;
    this.removeShadowEvent();

  }
```

joe-replin: This is great, thank you. I've made these suggested changes. 
Let's say you have a list of both resources & glossary buttons. Once you select either, the view changes and the top-level role remains while an additional list is included as a child within either resource or glossary view. I was thinking of adding a remove attribute in the resources extension but I don't know if you would prefer that kind of symbiotic relationship between the drawer & resources.

oliverfoster: There's almost certainly a part of the lifecycle in the drawer where that can be performed. Resources should not change drawer.

oliverfoster:
```js
  openCustomView(view, hasBackButton) {
    this.$('.js-drawer-holder').removeAttr('role');
    this._hasBackButton = hasBackButton;
    this._isCustomViewVisible = true;
    Adapt.trigger('drawer:empty');
    this.showDrawer();
    this.$('.drawer__holder').html(view);
  }
```
Removing the role in the `openCustomView` function works.

oliverfoster:
superceded by https://github.com/adaptlearning/adapt-contrib-core/pull/154

PR #115, opened 2022-05-13
issue/114 Added notify.isOpen

### LINKED ISSUE #114: Notify.isOpen
Add a getter to https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/views/notifyView.js, to check if a notify is currently open:
```javascript
get isOpen() {
  return this.stack.length > 0;
}
```

oliverfoster:
fixes #114 

### Added
* `notify.isOpen` to signify if any notifies are open

### DIFF
--- a/js/views/notifyView.js
+++ b/js/views/notifyView.js
@@ -25,6 +25,10 @@ export default class NotifyView extends Backbone.View {
   get stack() {
     return this._stack;
   }
+  
+  get isOpen() {
+    return (this.stack.length > 0);
+  }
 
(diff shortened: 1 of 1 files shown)


danielghost (approved):
As a preference, I tend to group methods together within files, e.g. getters, listeners etc., but this does what is needed. There are different styles for how people structure files anyway, so not going to enforce my tendencies.

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

Issue #102, opened 2022-04-25
Headings use displayTitle instead of title


oliverfoster:
https://github.com/adaptlearning/adapt-contrib-core/blob/master/templates/heading.hbs#L8-L15

oliverfoster:
Allow display title to come out if not display title is specified.

joe-replin:
Removing issue as current implementation allows for more flexible styling of display titles.

PR #116, opened 2022-05-13
issue/#102 - Headings use displayTitle instead of title

joe-replin:
Adjusts so the aria-label for all headings match what is visually represented on-screen. In relation to #102.


joe-replin:
Removing PR as current implementation allows for more flexible styling of display titles.

PR #118, opened 2022-05-14
issue/117 Added _shouldStoreAttempts

### LINKED ISSUE #117: Allow attemptsState to be turned off per model
To facilitate external branching blocks, the original block children should not store their attempts history as it is used by branching to save/restore the branching scenario.

oliverfoster:
fixes #117 

### Added
* `_shouldStoreAttempts` to prevent individual questions attempting to add to their attempt history


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):

Issue #124, opened 2022-05-24
Menu locking not working when using grouped items


joe-allen-89:
Original issue https://github.com/adaptlearning/adapt-contrib-boxMenu/issues/113

oliverfoster:
issue retracted

PR #125, opened 2022-05-24
issue/124 Locking updated to check _type on menu locking

joe-allen-89:
Fixes https://github.com/adaptlearning/adapt-contrib-core/issues/124

Fixed
Menu locking not working when using grouped items

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -715,7 +715,9 @@ export default class AdaptModel extends LockingModel {
   }
 
   checkLocking() {
-    const lockType = this.get('_lockType');
+    const lockType = (this.get('_type') !== 'course')
+      ? this.findAncestor().get('_lockType')
+      : this.get('_lockType');
 
     if (!lockType) return;
 
@@ -738,7 +740,9 @@ export default class AdaptModel extends LockingModel {
   }
 
   setSequentialLocking() {
-    const children = this.getAvailableChildModels();
+    const children = (this.get('_type') !== 'course')
+      ? this.findAncestor().findDescendantModels('pages', { where: { _isAvailable: true } })
+      : this.findDescendantModels('pages', { where: { _isAvailable: true } });
     // Start from second child
     children.slice(1).forEach((child, index) => {
       const previousChild = children[index];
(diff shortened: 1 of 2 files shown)


js/models/adaptModel.js
```diff
@@ -753,21 +757,27 @@ export default class AdaptModel extends LockingModel {
   }
 
   setUnlockFirstLocking() {
-    const children = this.getAvailableChildModels();
+    const children = (this.get('_type') !== 'course')
```
eleanor-heath, line 760: you could put the below into its own method and name the method `getChildren` to avoid code duplication

    ```
const children = (this.get('_type') !== 'course')
      ? this.findAncestor().findDescendantModels('pages', { where: { _isAvailable: true } })
      : this.findDescendantModels('pages', { where: { _isAvailable: true } });
```

**Review by eleanor-heath — APPROVED** [review]

**Comment by oliverfoster** [comment]
This seems like a very complex change to what was previously a very simple behaviour. 

The locking mechanism on each model, of any type (inc course, menu, page, article and block), was only interacting with its direct children. 
With this pr the root 'course' model now looks at all descendant pages only, but any other model looks at all of its parent's descendant pages. menu, page, article and block locking are excluded.

At alternative route, could be to introduce a new _lockType called, something like, 'sequentialFromPreviousCousin' which, [instead of skipping the first child](https://github.com/adaptlearning/adapt-contrib-core/blob/1f3907e59b82878a409568bf8c06c09d349b3cca/js/models/adaptModel.js#L742-L743), would lock the first child based upon the completion of the previous cousin?

```js 
const children = this.getAvailableChildModels();
const firstChild = children[0];
const firstChildType = firstChild.get('_type');
const previousCousin = this.findRelativeModel(`@${firstChildType}-1`);
if (previousCousin) {
   const isLockedByPreviousCousin = previousCousin.get('_isLocked') ||
        (
          !previousCousin.get('_isComplete') &&
          !previousCousin.get('_isOptional')
        );
   firstChild.set('_isLocked', isLockedByPreviousCousin);
}
// Start from second child
children.slice(1).forEach((child, index) => {
  const previousChild = children[index];
  // If previous was locked, all subsequent will be locked.
  const isLockedByPreviousChild = previousChild.get('_isLocked') ||
    (
      !previousChild.get('_isComplete') &&
      !previousChild.get('_isOptional')
    );
  child.set('_isLocked', isLockedByPreviousChild);
}, false);
```

That way, the locking stays on model children, for any locking and model type, we get a separate locking type for when we want the locking to continue forward from the previous cousin. i.e. if set on a menu, the first page of that menu would be locked by the previous page in the course, the last page of the previous menu. If set on a page, the first article in the page would be locked by the last article in the previous page, etc.

oliverfoster (changes requested):
As commented

chris-steele:
My advice on this would be the following:

1. Keep the existing behaviours as they are simple, understood and accepted practice.
2. Add code to permit truly arbitrary locking via the `_lockedBy` property: when a model with a `_lockedBy` list has **no ancestor** with a `_lockType` property watch the listed models and (un)lock the model as appropriate. This way any model can be locked on the completion of any other model(s).

oliverfoster:
@chris-steele isn't that exactly how custom locking works?

chris-steele:
Yes `custom` does this, what I should have added is that you need to use provide a `_lockType:"custom"` property at every level if you have a multi-tiered menu structure. Not a big issue, I just envisaged a slightly simpler configuration route.

PR #127, opened 2022-05-26
issue/126 Fixes notify accessibility blocking

### LINKED ISSUE #126: Notify accessibility blocking applies to wrong container
https://github.com/adaptlearning/adapt-contrib-core/blob/d5001ce2f5518412510b0e664923df7d3a96da42/js/views/notifyPopupView.js#L171

Should be applying to `this.$el` instead, allowing other content to be injected into the notify and be available to the screen reader.

oliverfoster:
fixes #126 

Allows for content to be injected into the dialog, outside of the content, and still be accessible

### Changed
* Block accessibillity from `.notify__popup` parent, the dialog container, rather than the `.notify__popup` as the dialog content

### DIFF
--- a/js/views/notifyPopupView.js
+++ b/js/views/notifyPopupView.js
@@ -168,7 +168,7 @@ export default class NotifyPopupView extends Backbone.View {
     $.inview();
     this.hasOpened = true;
     // Allows popup manager to control focus
-    a11y.popupOpened(this.$('.notify__popup'));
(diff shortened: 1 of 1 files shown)


eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

chris-steele (approved):

PR #128, opened 2022-05-26
issue/122 Fix orientation reporting

### LINKED ISSUE #122: orientation class update
The orientation class is currently only updated when the screensize breakpoint changes, which means an orientation selector can be inaccurate if the browser is resized on desktop.

If moving the orientation code to before https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/device.js#L139, should an orientation change trigger `device:changed`?

oliverfoster:
fixes #122 

### Fixed
* Orientation classes are now checked on ever resize
* Apple device orientation is calculated on the mismatch of the ratios between the screen and window, apple devices do not rotate the screen height and width on orientation changes but they do report window height and width orientation changes.
* Use window.innerWidth instead of window.screen.width for size calculation as screen.width doesn't account for pixel density

### DIFF
--- a/js/device.js
+++ b/js/device.js
@@ -71,7 +71,7 @@ class Device extends Backbone.Controller {
       : screenSizeConfig.small;
 
     const fontSize = parseFloat($('html').css('font-size'));
-    const screenSizeEmWidth = (this.screenWidth / fontSize);
+    const screenSizeEmWidth = (window.innerWidth / fontSize);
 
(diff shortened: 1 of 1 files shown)


eleanor-heath (approved):

guywillis (approved):
👍 

Tested:
* iPhone SE v15.4
* MacBook Pro Monterey v12.4 Safari v15.5

PR #141, opened 2022-06-16
Converted notify button divs to spans

guywillis:
Resolves https://github.com/adaptlearning/adapt-contrib-core/issues/139

* Converted button divs to spans


cahirodoherty-learningpool (approved):
👀

PR #146, opened 2022-06-21
issue/145 Automated aria levels

### LINKED ISSUE #145: Sequential Aria Level Headers
### Bug Description:

Aria level headers get skipped in numbering when article headers & block headers are not present. 

### Steps to Reproduce:

Using a screen reader, navigate through the headings, noting the heading level of each one.

### Actual Behavior:

When article & block headers are not present, heading levels will start at 1 for a page header and move to 4 for subsequent headers, with no mention of 2 or 3. Adapt currently assigns heading levels to the following elements:

- heading level 1 for page headers
- heading level 2 for article headers
- heading level 3 for block headers
- heading level 4 for component headers
- heading level 5 for component item headers

### Expected Behavior:

From https://www.w3.org/WAI/tutorials/page-structure/headings/
Skipping heading ranks can be confusing and should be avoided where possible: Make sure that a h2 is not followed directly by an h4, for example. It is ok to skip ranks when closing subsections, for instance, a h2 beginning a new section, can follow an h4 as it closes the previous section.

This would mean Adapt dynamically assigning aria heading levels to each level. A component would adopt a heading level of 2 if an article or block header did not currently exist.

oliverfoster:
fixes #145 

This is backward compatible, using the current defaults will produce the same outcome. Narrative and accordion will need to be updated to pass an id through for automated calculations.

### Current defaults
These are static unless overridden by `_ariaLevel` on the content.
```json
    "_accessibility": {
        "_ariaLevels": {
            "_menu": 1,
            "_menuGroup": 2,
            "_menuItem": 2,
            "_page": 1,
            "_article": 2,
            "_block": 3,
            "_component": 4,
            "_componentItem": 5,
            "_notify": 1
        }
    }
```

### Possible defaults
These aria levels are calculated dynamically, based upon the existence of a title on the defined parent and the offset through the tree, unless overridden in the tree by `_ariaLevel`.
```json
    "_accessibility": {
        "_ariaLevels": {
            "_menu": "1",
            "_menuGroup": "@menu+1",
            "_menuItem": "@menu+1",
            "_page": "1",
            "_article": "@page+1",
            "_block": "@article+1",
            "_component": "@block+1",
            "_componentItem": "@component+1",
            "_notify": 1
        }
    }
```
Examples:
If a component has no article or block title, its aria level will be calculated relative to the page heading level `1`, this is as the block and article offsets are not applied.
If a block has an `_ariaLevel` override, the components will increment from the block override.
`_ariaLevel` can be specified on any of the content models to override its heading level, this will then be used as the value for the heading level calculations for all descendants.

### Changed
* `a11y.ariaLevel` from flat arguments to object arguments, with backward compatibility for flat arguments, added `id` requirement to facilitate aria level calculations in the hierarchy

### Added
* `a11y.ariaLevel` calculates the aria level if given an id and a relative descriptor `@block+1`
* `a11y_aria_level` handlebars helper 

### Deprecated
* `a11y_attrs_heading` in favour of `a11y_aria_level` as it's clearer for the handlebars

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -200,7 +200,7 @@ class A11y extends Backbone.Controller {
       const isNumber =  !isNaN(level);
       const isTypeName = /[a-zA-z]/.test(level);
       if (!isTypeName && isNumber) {
-        // if an absolute value is found, use it, adding the accumilated offset
+        // if an absolute value is found, use it, adding the accumulated offset
         return parseInt(level) + offset;
       }
       // parse the level value as a relative string
@@ -209,7 +209,7 @@ class A11y extends Backbone.Controller {
       const nextLevel = ariaLevels?.['_' + relativeDescriptor.type];
       const hasModelId = Boolean(id);
       if (!hasModelId) {
-        logging.warnOnce('Cannot calculate appropraite heading level, no model id was specified');
+        logging.warnOnce('Cannot calculate appropriate heading level, no model id was specified');
         return calculateLevel(id, nextLevel, offset + relativeDescriptor.offset);
       }
       // try to find the next relevant ancestor, or use the specified model
@@ -218,10 +218,10 @@ class A11y extends Backbone.Controller {
       // check overrides, check title existence, adjust offset accordingly
       const hasNextTitle = Boolean(nextModel.get('displayTitle'));
       const nextModelOverride = nextModel.get('_ariaLevel');
-      const accumilatedOffset = offset + (hasNextTitle ? relativeDescriptor.offset : 0);
+      const accumulatedOffset = offset + (hasNextTitle ? relativeDescriptor.offset : 0);
       const resolvedLevel = nextModelOverride ?? nextLevel;
       // move towards the parents until an absolute value is found
-      return calculateLevel(nextModelId, resolvedLevel, accumilatedOffset);
+      return calculateLevel(nextModelId, resolvedLevel, accumulatedOffset);
     }
     return calculateLevel(id, override ?? level)
   }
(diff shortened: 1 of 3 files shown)


js/a11y.js
```diff
@@ -163,32 +164,66 @@ class A11y extends Backbone.Controller {
...
-
-    // first check to see if the Handlebars context has an override
-    if (overrideLevelOrType) {
-      defaultLevelOrType = overrideLevelOrType;
+  ariaLevel({
+    id = null,
+    level = "1",
+    override = null
+  } = {}) {
+    if (arguments.length > 1) {
```
cahirodoherty-learningpool, line 184: Shouldn't this be 
```suggestion
    if (arguments.length === 2) {
```
if we are trying to catch old implementations?

oliverfoster: New implementations will have only one argument. Either should be fine?

cahirodoherty-learningpool (approved):
👀

joe-replin (approved):
Implemented into my current build and this is brilliant. Thank you.

js/helpers.js
```diff
@@ -318,16 +319,35 @@ const helpers = {
...
-    const level = a11y.ariaLevel(levelOrType, this._ariaLevel);
-    return new Handlebars.SafeString(' role="heading" aria-level="' + level + '" ');
+  a11y_attrs_heading(level) {
+    logging.deprecated('a11y_attrs_heading, please use a11y_aria_level or a11y.ariaLevel');
+    const resolvedLevel = a11y.ariaLevel({
+      id: this._id ?? Array.from(arguments).lastItem?.data?.root?._id,
+      level,
+      override: this._ariaLevel
+    });
+    return new Handlebars.SafeString(' role="heading" aria-level="' + resolvedLevel + '" ');
```
eleanor-heath, line 333: could use template literals here `` role="heading" aria-level="${resolvedLevel}" ``

eleanor-heath (approved):

Issue #151, opened 2022-06-23
Missing semantic differentiation between status and heading


joe-replin:
## Bug Description:
In the table of contents, the status text "Complete" and "Incomplete" is included in the lesson title without semantic differentiation. This can render understanding of the lesson title unclear for screen reader users.

## Steps to Reproduce:
Using a screen reader, open the course prototype.
Navigate through the course titles and note the description of each course.

## Actual Behavior:
```
<div id="co-100-heading" class="js-heading-inner" role="heading" aria-level="2">
  <span class="aria-label">
  Incomplete Standard experience
  </span>
</div>
```

## Expected Behavior:
Update the labels of the lessons to provide clear separation between the lesson title and its status. The text "complete" by itself at the beginning of the heading can still make the purpose of the heading misleading without additional context. Consider moving status text to the end of the heading with context as to what the status represents (ex"[Course Title], course completed")

joe-replin:
Reopening this after further conversations on this to finalize plans. Status indicators in the aria-label should come first. Additionally, without context as to what is the status is reporting on can cause confusion as to the purpose of the status indicators. It was agreed the best path forward was to present all aria-headers in the following format:

[context (if applicable)] [status]. [title]

PR #152, opened 2022-06-23
Adjusts context between header & status in aria-labels

joe-replin:
Solves for #151

### DIFF
--- a/templates/heading.hbs
+++ b/templates/heading.hbs
@@ -7,9 +7,9 @@
     {{#if _isOptional}}
       {{{compile title}}}
       {{else if _isComplete}}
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

chris-steele (approved):

templates/heading.hbs
```diff
@@ -7,9 +7,9 @@
     {{#if _isOptional}}
       {{{compile title}}}
       {{else if _isComplete}}
-      {{_globals._accessibility._ariaLabels.complete}} {{{compile title}}}
+      {{{compile title}}} ({{_type}} {{_globals._accessibility._ariaLabels.complete}})
```
oliverfoster, line 10: Do we really want to include the type? It's not translatable.

PR #155, opened 2022-06-29
Added browser support override ability to course config ('targets')

cahirodoherty-learningpool:
fixes [#3312](https://github.com/adaptlearning/adapt_framework/issues/3312)

Allow a 'target' setting on course config to override desired browser compatibility

### DIFF
--- a/schema/config.model.schema
+++ b/schema/config.model.schema
@@ -439,7 +439,7 @@
         "strictMode": {
           "type": "boolean",
           "default": true,
-          "required": true,
+          "required": false,
           "inputType": "Checkbox",
           "validators": [],
           "title": "Use strict mode?",
--- a/schema/config.model.schema
+++ b/schema/config.model.schema
@@ -443,7 +443,7 @@
           "inputType": "Checkbox",
           "validators": [],
           "title": "Use strict mode?",
-          "help": "Uses ES6 modules. If AMD-style module definitions are required for plugins, turn this off"
+          "help": "Strict mode drops suport for AMD-style module definitions. If these are still required for any plugins, turn this setting off"
         },
         "targets": {
           "type": "string",
--- a/schema/config.model.schema
+++ b/schema/config.model.schema
@@ -443,14 +443,14 @@
           "inputType": "Checkbox",
           "validators": [],
           "title": "Use strict mode?"
+        },
+        "targets": {
+          "type": "string",
+          "title": "Supported browsers override",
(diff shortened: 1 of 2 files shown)


oliverfoster:
targets needs to go inside "build"

oliverfoster:
legacy aat needs this line correcting https://github.com/adaptlearning/adapt_authoring/blob/bed716953161dad2ba0d2a076725eb31f25b983c/lib/outputmanager.js#L273

to 

```js
configJson.build = Object.assign({}, configJson.build, { includes: includes });
```

such that it's non-destructive

oliverfoster (approved):

eleanor-heath (approved):

schema/config.schema.json
```diff
@@ -405,6 +405,26 @@
...
+      "type": "object",
+      "title": "Support for setting the course build type",
+      "properties" : {
+        "strictMode": {
+          "type": "boolean",
+          "default": true,
+          "required": true,
+          "inputType": "Checkbox",
+          "validators": [],
+          "title": "Use strict mode?"
```
taylortom, line 419: Think we could also do with a quick `description` of what this does if possible

taylortom (changes requested):
Thanks for looking into this 😃 a few changes needed for the JSON schema.

schema/config.model.schema
```diff
@@ -442,7 +442,8 @@
           "required": true,
           "inputType": "Checkbox",
           "validators": [],
-          "title": "Use strict mode?"
+          "title": "Use strict mode?",
+          "help": "Uses ES6 modules. If AMD-style module definitions are required for plugins, turn this off"
```
oliverfoster, line 446: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

Not quite right. ES6 modules are used regardless of this switch.

schema/config.model.schema
```diff
@@ -443,7 +443,7 @@
           "inputType": "Checkbox",
           "validators": [],
           "title": "Use strict mode?",
-          "help": "Uses ES6 modules. If AMD-style module definitions are required for plugins, turn this off"
+          "help": "Strict mode drops suport for AMD-style module definitions. If these are still required for any plugins, turn this setting off"
```
oliverfoster, line 446: You can use both AMD and ES modules in adapt whether or not strict mode is enabled.

Strict mode improves performance by tightly restricting the declaration of variables inside code blocks and closures which may cause some older code to stop working. Please leave this at true unless you are experiencing issue with old plugins or third party libraries.

cahirodoherty-learningpool: Thanks Oli

taylortom (changes requested):
Very sorry to be a pain (missed this on the first round :flushed:) but can we either get rid of the required or the default value for `strictMode` in the JSON schema (not sure which is more appropriate)

taylortom (approved):

PR #161, opened 2022-07-01
Add _friendlyId field

### LINKED ISSUE #160: Add _friendlyId field to content schema
This is a new human-readable ID for content objects which will replace the standard `_id` field on course build (useful in instances where the `_id` must be specified in the config of an extension, and is otherwise broken when the course is copied/imported elsewhere).

taylortom:
Fixes #160 

This is a new human-readable ID for content objects which will replace the standard `_id` field on course build (useful in instances where the `_id` must be specified in the config of an extension, and is otherwise broken when the course is copied/imported elsewhere).

### DIFF
--- a/schema/content.schema.json
+++ b/schema/content.schema.json
@@ -10,6 +10,11 @@
       "type": "string",
       "title": "Type"
     },
+    "_friendlyId": {
+      "type": "string",
+      "title": "Friendly id",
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):

PR #164, opened 2022-07-05
issue/123 Moved skip navigation button onscreen

### LINKED ISSUE #123: navigation skip button screen reader focus
As the `nav__skip-btn` is positioned off screen, this can cause issues with screen reader focus when used in designs which control the scroll, rather than using the standard overflow scrolling. In these designs, the page will scroll to an undesired position when the screen reader focus is applied to this button.

Rather than positioning off screen, and amending the top position on :focus, the same effect can be achieved using opacity instead:
```less
&__skip-btn {
  position: absolute;
  left: 0;
  margin-top: 0;
  top: 0;
  opacity: 0;
  z-index: 100;
  .transition(opacity @duration ease-in-out);

  html:not(.has-accessibility) & {
    .u-display-none;
  }

  &:focus {
    opacity: 1;
  }
}
```

oliverfoster:
fixes #123

### Fixed
* Keep skip navigation button onscreen rather than moving it off top. This stops screen readers from scrolling up when the button receives focus

### DIFF
--- a/less/core/nav.less
+++ b/less/core/nav.less
@@ -18,17 +18,18 @@
   &__skip-btn {
     position: absolute;
     left: 0;
-    margin-top: 0;
-    top: -100%;
+    top: 0;
     z-index: 100;
-    .transition(top @duration ease-in-out);
 
     html:not(.has-accessibility) & {
       .u-display-none;
     }
 
-    &:focus {
-      top: 0;
+    &:not(:focus) {
+      height: 0;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

chucklorenz:
+1

joe-allen-89 (approved):
Tested locally with FW v5.20.1 working as expected.

PR #170, opened 2022-07-06
issue/169 router.navigateToElement jquery object fix

### LINKED ISSUE #169: router.navigateToElement should take a jquery object and a selector
Currently it only takes a selector

oliverfoster:
fixes #169 

### Fixed
* Allowed navigateToElement to take a jquery array


cahirodoherty-learningpool (approved):
👀 

Apologies about the re-review. I'd mistakenly thought the jquery version wasn't handled correctly, but I can see that it is

joe-allen-89 (approved):
Tested with FW v5.20.1 locally, working as expected and able to navigate using jquery object.

PR #176, opened 2022-07-15
New: `--adapt-viewport-height` to aid viewport height positioning (fixes #175)

### LINKED ISSUE #175: viewport height measurements
The `vh` measurement isn't always accurate for mobile browsers due to the way they use a constant viewport height value to account for the address bar UI, which will minimise for scrolling pages. This can result in incorrectly positioned elements which utilise this measurement.

A CSS custom property should be added to https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/device.js which is updated on window resize to allow elements to respond to changes in the `window.innerHeight`.

This will resolve issues such as:
https://github.com/cgkineo/adapt-scrollSnap/issues/17

danielghost:
fixes #175.

### New
* CSS properties for viewport height

Ref: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty

### DIFF
--- a/js/device.js
+++ b/js/device.js
@@ -10,6 +10,7 @@ class Device extends Backbone.Controller {
     this.touch = Modernizr.touchevents;
     this.screenWidth = this.getScreenWidth();
     this.screenHeight = this.getScreenHeight();
+    this.setViewportHeight();
     this.browser = (this.bowser.browser.name || '').toLowerCase();
     this.version = (this.bowser.browser.version || '').toLowerCase();
     this.OS = this.getOperatingSystem().toLowerCase();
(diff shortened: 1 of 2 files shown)


oliverfoster:
First distinctly not ie11 compatible pr. Good job mate.

js/device.js
```diff
@@ -99,6 +100,10 @@ class Device extends Backbone.Controller {
       : window.innerHeight || this.$window.height();
   }
 
+  setViewportHeight() {
+    document.documentElement.style.setProperty('--viewport-height', `${window.innerHeight}px`);
```
oliverfoster, line 104: Could you prepend this with `--adapt-viewport-height` as these are global to the browser.

cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

joe-allen-89 (approved):
Tested with FW v5.20.0 working as expected.

Issue #184, opened 2022-07-21
`findViewByModelId` error


danielghost:
If `Adapt.parentView` is undefined, `findViewByModelId` will error due to:
https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/data.js#L232

Needs changing to:
```JavaScript
if (model === Adapt?.parentView?.model) return Adapt.parentView;
```

PR #185, opened 2022-07-22
`findViewByModelId` error

danielghost:
Fix for #184.

### DIFF
--- a/js/data.js
+++ b/js/data.js
@@ -228,9 +228,7 @@ class Data extends AdaptCollection {
   findViewByModelId(id) {
     const model = this.findById(id);
     if (!model) return;
-
-    if (model === Adapt.parentView.model) return Adapt.parentView;
-
+    if (model === Adapt?.parentView?.model) return Adapt.parentView;
     const idPathToView = [id];
     const currentLocationId = location._currentId;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

js/data.js
```diff
@@ -228,9 +228,7 @@ class Data extends AdaptCollection {
   findViewByModelId(id) {
     const model = this.findById(id);
     if (!model) return;
```
oliverfoster, line 230:
```suggestion
    if (!model || !Adapt.parentView) return;
```

oliverfoster (changes requested):

js/data.js
oliverfoster: I think this is much clearer.

js/data.js
```diff
@@ -228,9 +228,7 @@ class Data extends AdaptCollection {
...
-
-    if (model === Adapt.parentView.model) return Adapt.parentView;
-
+    if (model === Adapt?.parentView?.model) return Adapt.parentView;
```
oliverfoster, line 231:
```suggestion
```

oliverfoster (approved):

eleanor-heath (approved):

PR #205, opened 2022-07-28
Update: _isPartlyCorrect and questionView:submitted (fixes #196)

### LINKED ISSUE #196: Minor enhancements for core question functionality
At the moment it can be a bit fiddly establishing whether a question is partly correct and listening for when a component is both submitted and marked (without the use of defer/delay etc). I'm currently listening to `change:feedbackMessage` but it would be nice to have something more semantic.

I'm therefore proposing adding some small enhancements:

1. adding a `questionView:submissionComplete` trigger for when marking, update attempts and feedback have all been updated
2. adding an `_isPartlyCorrect` boolean value to the model
3. resetting `_isCorrect` and `_isPartlyCorrect` when the reset button is clicked

Another option instead of `_partlyCorrect` would be adding a correct state enum to the model so we can simply `switch` a single value when operating on the result. I think I actually like that better although _isCorrect would of course have to remain for compatibility.

Am happy to go ahead and create the PR if people are happy with these proposed additions and would love to get some more thoughts/ideas.

oliverfoster:
fixes #196

### New
* Added model property `_isPartlyCorrect` and view event `questionView:submitted` to the question API

### DIFF
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -136,15 +136,11 @@ class QuestionModel extends ComponentModel {
 
   // Sets _isCorrect:true/false based upon isCorrect method below
   markQuestion() {
-
-    if (this.isCorrect()) {
-      this.set('_isCorrect', true);
-    } else {
-      this.set('_isCorrect', false);
-    }
-
+    this.set({
+      _isCorrect: this.isCorrect(),
+      _isPartlyCorrect: this.isPartlyCorrect()
+    });
     this.updateRawScore();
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
Working as expected with latest FW.

eleanor-heath (approved):

ethan-lp (approved):

PR #207, opened 2022-08-01
New: added tooltip API (fixes: #193)

### LINKED ISSUE #193: adapt_framework: add tooltip api
### Subject of the issue
image

It's often useful to be able to display tooltips. An API like notify would be useful perhaps?

chris-steele:
Fixes #193 

### New
* Tooltip API
* `a11y.isForcedFocus` to show if focus was assigned by accessibility library or user interaction

### Description
image
image

* Tooltips will show on mouseover and focus. 
* Tooltips will be read by a screenreader on mouseover, on focus the screen reader will read the target's aria label.
* The plan is to add tooltips to all of the navigation icons and hotgraphic pins and possibly any other icon only button with an `aria-label`.
* The course will show tooltips with a default of `"bottom middle right"`, hotgraphic will use `"bottom middle middle"`

### Testing
Example: https://oliverfoster.github.io/shared/tooltips/index.html#/id/co-05
Use hotgraphic pr: https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/240

### Main API
Register a tooltip:
```js
import tooltips from 'core/js/tooltips'

tooltips.register({
  _id: "hotgraphic-c-05-1",
  _isEnabled: true,
  _position: "top middle right",
  text: "Text when target is focused or hovered",
  disabledText:  "Text when target aria-disabled or .is-disabled and is focused or hovered"
})
```
Add it to a target in the DOM:
```html
<div data-tooltip-id="hotgraphic-c-05-1" />
```

It is possible to specify `"{{ariaLabel}}"` for the `"text"` and `"disabledText"` properties, this will use the target's `aria-label` attribute or `.aria-label` child text. With hotgraphic you may use `"{{title}}"`, or any other item property.

### How it works
`_position` property syntax:
`"_position" = <area> <side> <side arrow position> <flow>`

`<area> = [outside/inside] = default is "outside"`
`<side> =  [middle/top/bottom/left/right] = default is "middle"`
`<side arrow position> = [middle/start/end] = default is "middle"`
`<flow> =  [middle/top/bottom/left/right] = default is "middle"`
`[outside/inside] [middle/top/bottom/left/right] [middle/start/end] [middle/top/bottom/left/right]`

All of the parameters are optional. so you can leave out `<area>` and it'll assume you meant `"outside"`.
In the same way `""`, `"middle"`, `"middle middle"`, `"outside middle"`, `"outside"` and `"middle outside"` all mean `"outside middle middle middle"` (which happens to be an invalid position).

### Invalid positions
`outside middle <side arrow position> <flow>` is not valid, as middle is not a side on the outside, it will be corrected to `"top"` or `"left"` depending on the `<flow>` orientation and a warning will be presented in the console. `inside middle <side arrow position> <flow>` is valid, as the inside has a middle.

### Assumed defaults
`"outside"` `<side arrow positioning>` will automatically follow the `<flow>` direction if its `<side arrow positioning>` is unspecified but its `<flow>` is specified. Such that `"outside left top"` will become `"outside left start top"` and `"outside bottom left"` will become `"outside bottom start left"`. This means the the flow and arrow will to follow the same direction, down and down or up and up by default.
`"inside"` `<side arrow positioning>` will automatically oppose the `<flow>` direction if its `<side arrow positioning>` is unspecified but its `<flow>` is specified. Such that `"inside left top"` will become `"inside left end top"` and `"inside bottom left"` will become `"inside bottom end left"`. This means that the flow direction moves away from the edge, text will flow down from the top or the text will flow right from the left by default.

### Target, sides and arrow positions

image

image

image

### Target styles

When the target is relative/static:
image

When the target is fixed positioned:
image

Positioning inside the target:
image

### Axis, side and flow

Vertical axis `"top"`:
`<side> <flow> = "top middle"`
`<side> <flow> = "top left"`
`<side> <flow> = "top right"`
image

Vertical axis `"bottom"`:
`<side> <flow> = "bottom middle"`
`<side> <flow> = "bottom left"`
`<side> <flow> = "bottom right"`
image

Horizontal axis `"left "`:
`<side> <flow> = "left middle"`
`<side> <flow> = "left top"`
`<side> <flow> = "left bottom"`
image

Horizontal axis `"right"`:
`<side> <flow> = "right middle"`
`<side> <flow> = "right top"`
`<side> <flow> = "right bottom"`
image

Test render:
If `course.json:_tooltips._allowTest = true` and move over tooltip whilst holding cntrl.
image

### Algorithm 
Works in 4 stages:
0. Calculate the appropriate defaults and correct the `"_position"` attribute, showing warnings where necessary
1. Render as per `"_position"` config
2. If needed, swap sides, switch axis, fill width or fill height
3. If overflowing the sides, snap to the left, right, top and bottom sides as necessary. Content will be truncated if it is too much to fit on screen.

The algorithm will always prefer the sides as specified, except with encountering the viewport edges.
Very large tooltips will try to fill the largest available space.

All styling in done in less, classes and area sizes are calculated in the js and passed to the less with css variables.

### LESS variables
```
@tooltip-color: rgba(0,0,0,0.85); // background color, foreground is currently white
@tooltip-radius: 0.35rem; // max 1rem
@tooltip-offset: 0.35rem; // max 1rem, distance between arrow and end of tooltip edge
@tooltip-arrow-width: 0.7rem; // max 1rem
@tooltip-arrow-height: 0.35rem;
@tooltip-arrow: true; // turn off arrows completely
@tooltip-distance: 1rem; // tooltip distance from target 
@tooltip-viewport-padding: 8px; // edge padding
```

### CSS JavaScript calculated variables
This exist on the parent of each tooltip and are relative to the tooltip and target
```
// Derived from the less variables
  --adapt-tooltip-offset: @tooltip-offset;
  --adapt-tooltip-distance: @tooltip-distance;
  --adapt-tooltip-viewport-padding: @tooltip-viewport-padding;
// Target position
  --adapt-tooltip-target-position-left: 0px;
  --adapt-tooltip-target-position-top: 0px;
  --adapt-tooltip-target-position-right: 0px;
  --adapt-tooltip-target-position-bottom: 0px;
  --adapt-tooltip-target-position-width: 0px;
  --adapt-tooltip-target-position-height: 0px;
// Distance to viewport padding from target edges
  --adapt-tooltip-target-distancetoedge-left: 0px;
  --adapt-tooltip-target-distancetoedge-top: 0px;
  --adapt-tooltip-target-distancetoedge-right: 0px;
  --adapt-tooltip-target-distancetoedge-bottom: 0px;
// Browser scroll offsets
  --adapt-tooltip-scroll-top: 0px;
  --adapt-tooltip-scroll-left: 0px;
```
Classes
```
is-outside // tooltip outside target
is-inside 
is-vertical-axis // vertical properties represent side, horizontal properties represent flow
is-horizontal-axis
is-left
is-middle
is-right
is-top
is-center
is-bottom
is-fill-width // when horizontal axis: top or bottom of the arrow, fill width left or right of the target (quarter), when vertical axis: fill width full top or bottom (half)
is-fill-height
is-snap-left // make tooltip snap to viewport left
is-snap-right
is-snap-top
is-snap-bottom
is-arrow-start // put arrow at start of edge
is-arrow-middle
is-arrow-end
is-arrow-snap // make arrow snap to viewport edge
```

### Animations
There is a single small scale and opacity animation, transform origin is from the center of the target for outside positions. With inside positions, the transform origin is the appropriate edge or corner.


chris-steele:
Set up as PR for reviews, but this is just a first unseen draft

templates/tooltip.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { html, classes, compile } from 'core/js/reactHelpers';
```
cahirodoherty-learningpool, line 2: Minor: html doesn't need imported

js/tooltips.js
```diff
@@ -0,0 +1,153 @@
...
+      // find the 'corner' with the most space
+      const isTopPreferred = availableHeight - (targetBoundingRect.bottom + tooltipsHeight) < targetBoundingRect.top - tooltipsHeight;
+      const isLeftPreferred = availableWidth - (targetBoundingRect.right + tooltipsWidth) < targetBoundingRect.left - tooltipsWidth;
+
+      if (isTopPreferred) {
+        isLeftPreferred ? alignTopLeft() : alignTopRight();
+      } else {
+        isLeftPreferred ? alignBottomLeft() : alignBottomRight();
+      }
+    }
```
oliverfoster, line 141:
```suggestion
    if (canAlignBottomRight) return alignBottomRight();
    if (canAlignTopRight) return alignTopRight();
    if (canAlignBottomLeft) return alignBottomLeft();
    if (canAlignTopLeft) return alignTopLeft();

    // find the 'corner' with the most space
    const isTopPreferred = availableHeight - (targetBoundingRect.bottom + tooltipsHeight) < targetBoundingRect.top - tooltipsHeight;
    const isLeftPreferred = availableWidth - (targetBoundingRect.right + tooltipsWidth) < targetBoundingRect.left - tooltipsWidth;

    if (isTopPreferred && isLeftPrefered) return alignTopLeft();
    if (isTopPrefered) return alignTopRight();
    if (isLeftPrefered) return alignBottomLeft();
    alignBottomRight();
```

Move all positioning stuff to ToolTipView

js/tooltips.js
```diff
@@ -0,0 +1,153 @@
+import Adapt from 'core/js/adapt';
+import TooltipView from 'core/js/views/tooltipView';
```
oliverfoster, line 2: Please make sure the file names for tooltipView and tooltipModel are capitalised to denote that a class is returned.

oliverfoster (changes requested):

js/tooltips.js
chris-steele: i'll amend this once you've had chance to see the diffs

js/tooltips.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+    this._tooltipData.push(new TooltipModel(data));
+  }
+
+  show(tooltip) {
+    this.hide();
+
+    console.log('show tooltip', tooltip.get('_id'));
+    this.tooltipsView = new TooltipView({model:tooltip, target:this.$mouseoverEl});
+    this.tooltipsView.attach('appendTo', $('body'));
```
oliverfoster, line 52: This was good to live in the dom from the start as you had it before.
The reason is that `aria-live` regions must exists in the DOM and it is their content changes that cause the screen reader to read.

See the 'proceedure' sections of the spec https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA19#procedure

Perhaps hide and show the container instead of adding and removing it?

oliverfoster (changes requested):

oliverfoster (changes requested):
I have refactored in https://github.com/adaptlearning/adapt-contrib-core/pull/214 as I found the comments a little too complex to write down towards the end.

Let me know what you think, it retains all the essence but is less code where applicable.

oliverfoster:
Immediate bits that need doing
* Tooltips for fixed position elements
* _isEnabled schema för course.json
* Back button

js/tooltips.js
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * @param {Object} tooltipData
+   * @param {boolean} tooltipData._isEnabled
+   * @param {string} tooltipData._id Id of registered tooltip text
+   * @param {string} tooltipData.text Text to be displayed
+   */
+  register(tooltipData) {
+    if (!tooltipData._id) return logging.warn('Tooltip cannot be registered with no id');
+    const existingTooltip = this.getTooltip(tooltipData._id);
+    if (existingTooltip) return logging.warn(`Tooltip with id ${existingTooltip._id} already registered`);
```
oliverfoster, line 83: I've been thinking about this and translations. It's probably best to overwrite and existing tooltip as models will need to reregister on language change.

less/core/tooltip.less
```diff
@@ -0,0 +1,21 @@
+.tooltip {
+  display: none;
+  position: absolute;
```
oliverfoster, line 3: For fixed positioned tooltips this should be position:fixed and the top, left won't have a scroll offset applied.

You can figure out if the $mouseoverEl is fixed position using something like this:
```js
const isFixedPosition = Boolean($mouseoverEl.parents().add($mouseoverEl).filter((index, el) => $(el).css('position') === 'fixed').length);
```

schema/course.model.schema
```diff
@@ -348,6 +348,28 @@
                   "type": "number",
                   "title": "Navigation bar order",
                   "default": 0
+                },
+                "_navTooltip": {
```
oliverfoster, line 352: I guess you'd add something like `_globals._extension._navigation._backTooltip` for the back button?

js/views/TooltipView.js
```diff
@@ -0,0 +1,90 @@
...
+          };
+        }
+      }
+      // Bottom right, default
+      return {
+        left: `${targetBoundingRect.right + scrollLeft}px`,
+        top: `${targetBoundingRect.bottom + scrollTop}px`
+      };
+    }
+    this.model.set(getPosition());
```
oliverfoster, line 88: I guess you'd add the scrollLeft and scrollTop to the result of `getPosition` depending on `isFixedPosition`?

js/tooltips.js
```diff
@@ -0,0 +1,136 @@
...
+
+  getConfig() {
+    return Adapt.course.get('_tooltips');
+  }
+
+  registerGlobalTooltips() {
+    const tooltips = Adapt.course.get('_globals')?._tooltips;
+    if (!tooltips) return;
+    tooltips.forEach(tooltip => this.register(tooltip));
+  }
```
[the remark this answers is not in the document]
chris-steele: @oliverfoster can we maintain `_globals._tooltips` so that, for example, an arbitrary component instance can utilise the tooltip API without requiring change to the underlying plugin?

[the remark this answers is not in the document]
oliverfoster: Why not just update the plugin so that it works in the AAT? Not doing that would seem like diverting work from something worthwhile? It's just a tooltip registration call, so not a high barrier?

Do you have any examples where that wouldn't work?

less/core/tooltip.less
chris-steele: @oliverfoster if I apply position:fixed to a test element and mouseover it the position of the tooltip is correct. could you provide a quick example of why we need to make the suggested change?

chris-steele: Ah I see - scrolling while the tooltip it is visible - thanks!

js/drawer.js
```diff
@@ -18,6 +19,10 @@ Drawer.triggerCustomView = function(view, hasBackButton) {
 
 Adapt.on({
   'adapt:start'() {
+    tooltips.register({
+      _id: 'drawer',
+      ...Adapt.course.get('_globals')?._extensions?._drawer?._navTooltip || {}
+    });
```
oliverfoster, line 25: Might be best to move this to `app:languageChanged` below?

js/models/TooltipModel.js
```diff
@@ -0,0 +1,10 @@
+import Backbone from 'backbone';
+
+export default class TooltipModel extends Backbone.Model {
+
+  defaults() {
+    return {
+      text:''
```
oliverfoster, line 7:
```suggestion
      text:' '
```

js/models/TooltipModel.js
```diff
@@ -0,0 +1,12 @@
+import Backbone from 'backbone';
+
+export default class TooltipModel extends Backbone.Model {
+
+  defaults() {
+    return {
+      _id: null.
```
oliverfoster, line 7:
```suggestion
      _id: null,
```

js/models/TooltipModel.js
```diff
@@ -0,0 +1,12 @@
+import Backbone from 'backbone';
+
+export default class TooltipModel extends Backbone.Model {
+
+  defaults() {
+    return {
+      _id: null,
+      _isEnabled: true,
+      text: ''
```
oliverfoster, line 9:
```suggestion
      _classes: '',
      text: ''
```

chris-steele:
> Can you install ESLint please? You have a series of syntax errors that can be corrected.

I do have it installed and I have formatted so not sure why I'm getting different results to you?

oliverfoster:
Are you editing the core in an adapt_framework with npm install run? The styles don't get applied unless.

oliverfoster:
You'll need to register the back button tooltip from the navigation controller and update the handlebars to pull it through.

oliverfoster:
Awaiting styling

oliverfoster:
@kirsty-hames how's this coming along?

js/tooltips.js
```diff
@@ -0,0 +1,87 @@
...
+
+  getConfig() {
+    return Adapt.course.get('_tooltips');
+  }
+
+  attachToBody() {
+    this.tooltipsView = this.tooltipsView || new TooltipView();
+    const $el = this.tooltipsView.$el;
+    if ($el[0].parentNode && $el.is(':last-child')) return;
```
eleanor-heath, line 31: could `$el[0].parentNode` ever return undefined? Could guard it using the optional chaining operator ? `$el[0]?.parentNode` to make sure we dont get a type error of Cannot read properties of undefined

js/views/TooltipView.js
```diff
@@ -0,0 +1,96 @@
...
+  render() {
+    const Template = templates.tooltip;
+    ReactDOM.render(<Template {...this.model.toJSON()} />, this.el);
+  }
+
+  position($mouseoverEl) {
+    const targetBoundingRect = this.$target[0].getBoundingClientRect();
+
+    const availableWidth = $('html')[0].clientWidth;
+    const availableHeight = $('html')[0].clientHeight;
```
eleanor-heath, line 47: can add the optional chaining operator to the available width and available height constants too

```suggestion
    const availableWidth = $('html')[0]?.clientWidth;
    const availableHeight = $('html')[0]?.clientHeight;
```

js/views/TooltipView.js
```diff
@@ -0,0 +1,96 @@
...
+    const tooltipsWidth = this.$('.tooltip').width();
+    const tooltipsHeight = this.$('.tooltip').height();
+
+    const canAlignBottom = targetBoundingRect.bottom + tooltipsHeight < availableHeight;
+    const canAlignRight = targetBoundingRect.right + tooltipsWidth < availableWidth;
+    const canAlignBottomRight = canAlignBottom && canAlignRight;
+
+    const isFixedPosition = Boolean($mouseoverEl.parents().add($mouseoverEl).filter((index, el) => $(el).css('position') === 'fixed').length);
```
eleanor-heath, line 55: this line is very long, could add a new line for each function that is chained
```suggestion
    const isFixedPosition = Boolean($mouseoverEl.parents()
        .add($mouseoverEl)
        .filter((index, el) => 
                $(el).css('position') === 'fixed'
        ).length);
```

js/views/TooltipView.js
```diff
@@ -0,0 +1,96 @@
...
+    const scrollOffsetTop = isFixedPosition ? 0 : $(window).scrollTop();
+    const scrollOffsetLeft = isFixedPosition ? 0 : $(window).scrollLeft();
+
+    this.model.set('_isFixedPosition', isFixedPosition);
+
+    function getPosition() {
+      if (!canAlignBottomRight) {
+        // Find the 'corner' with the most space from the viewport edge
+        const isTopPreferred = availableHeight - (targetBoundingRect.bottom + tooltipsHeight) < targetBoundingRect.top - tooltipsHeight;
+        const isLeftPreferred = availableWidth - (targetBoundingRect.right + tooltipsWidth) < targetBoundingRect.left - tooltipsWidth;
```
eleanor-heath, line 65: These 2 lines also very long - can we extract some of the calculations into variables?

js/views/TooltipView.js
```diff
@@ -0,0 +1,96 @@
...
+            top: `${targetBoundingRect.bottom + scrollOffsetTop}px`
+          };
+        }
+      }
+      // Bottom right, default
+      return {
+        left: `${targetBoundingRect.right + scrollOffsetLeft}px`,
+        top: `${targetBoundingRect.bottom + scrollOffsetTop}px`
+      };
+    }
```
eleanor-heath, line 93: A lot of duplication in this, for example `${targetBoundingRect.bottom + scrollOffsetTop}px` is used twice, `${targetBoundingRect.top - tooltipsHeight + scrollOffsetTop}px` is used twice and `${targetBoundingRect.right + scrollOffsetLeft}px` is used twice and `${targetBoundingRect.left - tooltipsWidth + scrollOffsetLeft}px` is used twice. Can we use variables here to reduce duplication?

less/core/tooltip.less
```diff
@@ -0,0 +1,25 @@
...
+  top:0;
+  left:0;
+  background-color: black;
+  color: white;
+  overflow: visible;
+  min-width: 100px;
```
eleanor-heath, line 9: should we use rem rather than px?

StuartNicholls: Hey Eleanor, thanks for your feedback on this, I've refactored most of this, Ill push it up next week but will make sure I capture all your comments above!

js/views/TooltipView.js
StuartNicholls: The refactor I've done adds a fair bit of calculations and gets a bit complex in places. My feeling is single lines might be more readable overall but will have a look

StuartNicholls:
Updates
======

- Added styling to tooltips
- Added methods so css styling / animation can alter the appearance (rather than compex jSON) without breaking the functionality (within reason!)
- Added styling hooks to vanilla for customisation
- Added button targeting for tooltips (middle is default additionally start / end , more can be added if/as necessary)
- Added tooltip position style ( left / middle / right & top / bottom )
- Added screen edge detection for button targeting (including exclusion zone)
- Improved screen edge detection for tooltip position style (including exclusion zone)
- Added fallback so if the tooltip is large and wont fit on screen, it resizes and moves the tooltip arrow to the correct position. 
- Adjusted method for speech so tooltip text is read *after* any aria labels are announced on the button rather than interrupting it
- Added method to apply a tooltip when the button is disabled (also works on touch devices) 

 

Proposals to consider
================

JSON configuarable distance from buton?
Adjustable exclusion border?

Proposed field names in JSON. I'd like some specific feedback here, as not sure what to call theses. positioning/placement will then need to be added to the hotgraphic schema

```
  "_tooltip": {
    "_isEnabled": true,
    "_placement": "middle", // optional
    "_position": "top right", // optional
    "text": "Tooltip text here",
    "disabledText": "This button is disabled" // optional
  }
```

chucklorenz:
@StuartNicholls 
"Added button targeting for tooltips (middle is default additionally start / end , more can be added if/as necessary)"
""_placement": "middle", // optional"
Do these refer to what part of the button the tooltip arrow is pointing to?
If so, maybe something like _arrowTarget

chucklorenz:
@StuartNicholls 
I'm a little confused, But I acknowledge that you might have enough people around you who can brainstorm an adequate attribute name without me getting involved. So need to respond if it gets too complicated.
(Yet on the other hand, We're going to need explanations in order to know how to use it...)
I'm attaching an image in case it helps your discussions.

   
tooltip

StuartNicholls:
> @StuartNicholls I'm a little confused, But I acknowledge that you might have enough people around you who can brainstorm an adequate attribute name without me getting involved. So need to respond if it gets too complicated. (Yet on the other hand, We're going to need explanations in order to know how to use it...) I'm attaching an image in case it helps your discussions.
> 
> tooltip

Thanks Chuck, Id recommend having a play with it, its easier to visualise. I had intended to work on a wiki page but Im off on holiday until the 11th so wanted to get it up before then. 

It is slightly complicated.. I agree!

oliverfoster:
It appears that `_placement` values of `start` and `end` do the same thing @chucklorenz and that it's not clear what it's meant to do.
`_position` is clearer, it's effectively a combination of `top left` etc, to denote the prefered position of the tooltip relative to the button, as in your 1-8 diagram.
I'm just working on refactoring the code to make it a little clearer.

StuartNicholls:
yep, _placement start & end align the tooltip to the edge of the button, I toyed with the idea of just calling this target 'edge'. But for the moment, I've followed the same pattern as most of the popular position engines for tooltips although there's slightly more flexibility here with whats called  _position which is slightly confusing the issue.

oliverfoster:
The refactored works as below:

`_position = <area> <arrow position> <side> <flow>`

`<area> = [outside/inside] = default is outside`
`<arrow position> = [middle/start/end] = default is middle`
`<side> =  [middle/top/bottom/left/right] = default is middle`
`<flow> =  [middle/top/bottom/left/right] = default is middle`

`[outside/inside] [middle/start/end] [middle/top/bottom/left/right] [middle/top/bottom/left/right]`

image

image

image


When the target is relative/static:
image

When the target is fixed positioned:
image

Positioning inside the target:
image

`<side> <flow> = "top middle"`
`<side> <flow> = "top left"`
`<side> <flow> = "top right"`
image

`<side> <flow> = "bottom middle"`
`<side> <flow> = "bottom left"`
`<side> <flow> = "bottom right"`
image

`<side> <flow> = "left middle"`
`<side> <flow> = "left top"`
`<side> <flow> = "left bottom"`
image

`<side> <flow> = "right middle"`
`<side> <flow> = "right top"`
`<side> <flow> = "right bottom"`
image

Algorithm works in 4 stages:
1. Render as per `position` config

oliverfoster:
Ready.

Use hotgraphic branch issue/239: https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/240

Course.json:
```
  "_tooltips": {
    "_isEnabled": true,
    "_position": "middle bottom right",
    "_allowTest": true
  },
```
Test render can be performed by holding ctrl when mouse over. It will show all tooltip edge styles.

components.json:
```
{
        "_id": "c-35",
        "_parentId": "b-05",
        "_type": "component",
        "_component": "hotgraphic",
        "_classes": "",
        "_layout": "full",
        "_comment": "setCompletionOn = inview | allItems",
        "_setCompletionOn": "allItems",
        "_canCycleThroughPagination": false,
        "_hidePagination": false,
        "_isNarrativeOnMobile": true,
        "_useNumberedPins": false,
        "_useGraphicsAsPins": false,
        "title": "Hot Graphic",
        "displayTitle": "Hot Graphic",
        "body": "You can add interactivity to an image by using the <b>Hot Graphic</b>. This component allows you to position icons over an image. When an icon is selected, content associated with its corresponding location is displayed in a window over the image. This component will fall back to a <b>Narrative</b> when viewed on mobile.<br><br>This component is always spanned.",
        "instruction": "Select the icons to find out more.",
        "mobileInstruction": "Select the plus icon followed by the next arrow to find out more.",
        "_graphic": {
            "src": "course/en/images/hotgraphic.png",
            "alt": "alt text"
        },
        "_items": [
            {
                "title": "Hot Graphic item 1 ",
                "body": "This is display text associated with item 1.",
                "_graphic": {
                    "src": "course/en/images/single-width.png",
                    "alt": "",
                    "_classes": ""
                },
                "strapline": "Hot Graphic strapline 1",
                "_classes": "",
                "_top": 42,
                "_left": 8.5,
                "_tooltip": {
                    "_isEnabled": true,
                    "disabledText": "This is disabled",
                    "text": "(course default) {{ariaLabel}}"
                }
            },
            {
                "title": "Hot Graphic item 1 ",
                "body": "This is display text associated with item 1.",
                "_graphic": {
                    "src": "course/en/images/single-width.png",
                    "alt": "",
                    "_classes": ""
                },
                "strapline": "Hot Graphic strapline 1",
                "_classes": "",
                "_top": 62,
                "_left": 8.5,
                "_tooltip": {
                    "_isEnabled": true,
                    "disabledText": "This is disabled",
                    "text": "(course default) Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1."
                }
            },
            {
                "title": "Hot Graphic item 1 ",
                "body": "This is display text associated with item 1.",
                "_graphic": {
                    "src": "course/en/images/single-width.png",
                    "alt": "",
                    "_classes": ""
                },
                "strapline": "Hot Graphic strapline 1",
                "_classes": "",
                "_top": 30,
                "_left": 8.5,
                "_tooltip": {
                    "_isEnabled": true,
                    "disabledText": "This is disabled",
                    "text": "(left bottom) {{ariaLabel}}",
                    "_position": "left bottom"
                }
            },
            {
                "title": "Hot Graphic item 1 ",
                "body": "This is display text associated with item 1.",
                "_graphic": {
                    "src": "course/en/images/single-width.png",
                    "alt": "",
                    "_classes": ""
                },
                "strapline": "Hot Graphic strapline 1",
                "_classes": "",
                "_top": 10,
                "_left": 80,
                "_tooltip": {
                    "_isEnabled": true,
                    "disabledText": "This is disabled",
                    "text": "(right middle) {{ariaLabel}}",
                    "_position": "right middle"
                }
            },
            {
                "title": "Hot Graphic item 2 ",
                "body": "This is display text associated with item 2.",
                "_graphic": {
                    "src": "course/en/images/single-width.png",
                    "alt": "",
                    "_classes": ""
                },
                "strapline": "Hot Graphic strapline 2",
                "_classes": "",
                "_top": 62,
                "_left": 26.5,
                "_tooltip": {
                    "_isEnabled": true,
                    "text": "(left middle) Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1.",
                    "_position": "left middle"
                }
            },
            {
                "title": "Hot Graphic item 3 ",
                "body": "This is display text associated with item 3.",
                "_graphic": {
                    "src": "course/en/images/single-width.png",
                    "alt": "",
                    "_classes": ""
                },
                "strapline": "Hot Graphic strapline 3",
                "_classes": "",
                "_top": 62,
                "_left": 49,
                "_tooltip": {
                    "_isEnabled": true,
                    "text": "(course default) Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. Hot Graphic strapline 1. "
                }
            }
        ],
        "_pageLevelProgress": {
            "_isEnabled": true,
            "_isCompletionIndicatorEnabled": true
        }
    },
```

chucklorenz:
@oliverfoster 
3 or 4 parameters for "_position"?
  "_tooltips": {
    "_isEnabled": true,
    "_position": "&lt;area&gt; middle bottom right",
    "_allowTest": true
  },

oliverfoster:
> @oliverfoster 3 or 4 parameters for "_position"? "_tooltips": { "_isEnabled": true, "_position": "<area> middle bottom right", "_allowTest": true },

All of the parameters are optional. so you can leave out area and it'll assume you meant "outside".

Guy recommended I remove all the positioning stuff from the schemas, so I've added some defaults into the course and hotgraphic accordingly. "middle bottom middle" for hotgraphic, "middle bottom right" for course default. It is still possible to override using the json "`_position`" attribute.

chucklorenz:
gotcha, thank you (and you did mention default in your comment--just overlooked that)

chucklorenz:
I don't think the test is working for me.
image

oliverfoster:
It worked just fine on the button target. 

The reason it doesn't look like my test is that I haven't added the code for a tooltip on the background image, because I'm not intending to keep one there. This test has one there though https://oliverfoster.github.io/shared/tooltips/index.html

chucklorenz:
https://oliverfoster.github.io/shared/tooltips/index.html looks good. I'll try reinstalling to see if I can match that.

chucklorenz:
All installed and looking like https://oliverfoster.github.io/shared/tooltips/index.html.

Using this as a model:
`"_position" = <area> <arrow position> <side> <flow>`

I tried changing the arrow position to "start" and got the following results which I found unexpected. Are these correct?

1a
1ac

-----

1b
1bc

-----

1c
1cc

chucklorenz:
What should I be focusing on in the latter 2 images, the ones with values "right middle" and "right bottom"? I see a change in the arrow position but I expected that to be governed by "start"

oliverfoster:
The arrow position is relative to the side, not the tooltip. The tooltip flow is relative to the arrow.

chucklorenz:
>Interesting thought though... the flow direction is effectively just an arrow position relative to the tooltip, inverted. If I inverted flow we'd have something we could describe as <side> <side arrow position> <tooltip arrow position>, which would be what you described, I think, is that easier to think about?

"Easier"? Not for me. Easiest for me is to think of 2 objects: button and tooltip (arrow is part of tooltip). I work first with the button by choosing a side. Then I work with the tooltip by aligning it relative to an axis (flow) and then by adjusting its arrow. Anyway, that's how my mind works 🤷🏼‍♂️

oliverfoster:
Yes, indeed, I'll have a look and ask Stu. I effectively expanded on what he'd written, I'm hoping I haven't made a wild deviation.

> I work first with the button by choosing a side. Then I work with the tooltip by aligning it relative to an axis (flow) and then by adjusting its arrow..

Yes... That's where we're getting muddled. The side has an arrow position, not the tooltip in this implementation.

oliverfoster:
> * I'm only seeing arrows on items that include "_position". Intentional?

There should be default position in the latest commit to hotgraphic. "Inside" tooltips do not have arrows. Including "_position" should not be necessary.

> * With the <side> and <flow> using the same values and <arrow position> also using 'middle', how would this be interpretted: "_position": "middle top"? As "middle top middle" or "middle middle top"?

As "middle top middle". It assumes the vertical axis label top is the side.

chucklorenz:
Maybe others caught on quicker than I, but it makes sense to me now that the arrow position is relative to the button. Your test example, Ollie, with lots and lots of text demonstrates how impractical it would be to base the position of the arrow on the "start", "middle", or "end" of the oversized tooltip itself.

oliverfoster:
Ok, I'm happy with this now. It seems much less confusing and much more useful.

swashbuck:
@chris-steele @oliverfoster How difficult would it be to add an option to _always_ show a particular tooltip? For a project, we need to show text labels for Hot Graphic pins, and they need to always be visible. I thought I could use the [Hot Graphic PR](https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/240), but the tooltips only appear on mouseover and focus.

Would it be better to create a separate Hot Graphic issue for static labels, or can the tooltip API be modified to always show specific (not global) tooltips?

oliverfoster:
> @chris-steele @oliverfoster How difficult would it be to add an option to _always_ show a particular tooltip? For a project, we need to show text labels for Hot Graphic pins, and they need to always be visible. I thought I could use the [Hot Graphic PR](https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/240), but the tooltips only appear on mouseover and focus.
> 
> Would it be better to create a separate Hot Graphic issue for static labels, or can the tooltip API be modified to always show specific (not global) tooltips?

It wouldn't be at all difficult I wouldn't have thought. A few changes needed. Separate to this pr though. Does this one work as expected otherwise?

swashbuck (approved):
👍

cahirodoherty-learningpool (approved):
👀

guywillis:
Tooltips look to be a very nice feature - thank you to all those who have contributed.

The majority of the required information is defined above and I'm able to follow the logic. 

One area that I didn't see defined (sorry if I've missed it!) was the defined functionality for touch devices. I've noted the functionality below, just wanted to check to see if that was the expectation.

If I open and close the drawer using the buttons provided (navigation drawer and drawer close buttons) I am shown the tooltip after closing. This tooltip persists whilst I scroll until I manually press the screen again (likely when trying to select an interaction).

IMG_0001

Similarly, if I select a hotspot within the hotgraphic interaction and close the popup using the subsequent notify close button the tooltip is present until I press the screen or interaction.

IMG_0002

For reference the json properties I have been using to test:

```
    "_extensions": {
      "_drawer": {
        "_navOrder": 100,
        "_navTooltip": {
          "_isEnabled": true,
          "text": "{{ariaLabel}}"
        }
      },
      "_navigation": {
        "_skipButton": {
          "_navOrder": -100
        },
        "_backButton": {
          "_navOrder": 0
        },
        "_backNavTooltip": {
          "_isEnabled": true,
          "text": "{{ariaLabel}}"
        },
        "_spacers": [
          {
            "_navOrder": 0
          }
        ]
      }
    }
  },
  "_tooltips": {
    "_isEnabled": true
  },
```

oliverfoster:
> Tooltips look to be a very nice feature - thank you to all those who have contributed.
> 
> The majority of the required information is defined above and I'm able to follow the logic.
> 
> One area that I didn't see defined (sorry if I've missed it!) was the defined functionality for touch devices. I've noted the functionality below, just wanted to check to see if that was the expectation.
> 
> If I open and close the drawer using the buttons provided (navigation drawer and drawer close buttons) I am shown the tooltip after closing. This tooltip persists whilst I scroll until I manually press the screen again (likely when trying to select an interaction).
> 
> IMG_0001
> 
> Similarly, if I select a hotspot within the hotgraphic interaction and close the popup using the subsequent notify close button the tooltip is present until I press the screen or interaction.
> 
> IMG_0002
> 
> For reference the json properties I have been using to test:
> 
> ```
>     "_extensions": {
>       "_drawer": {
>         "_navOrder": 100,
>         "_navTooltip": {
>           "_isEnabled": true,
>           "text": "{{ariaLabel}}"
>         }
>       },
>       "_navigation": {
>         "_skipButton": {
>           "_navOrder": -100
>         },
>         "_backButton": {
>           "_navOrder": 0
>         },
>         "_backNavTooltip": {
>           "_isEnabled": true,
>           "text": "{{ariaLabel}}"
>         },
>         "_spacers": [
>           {
>             "_navOrder": 0
>           }
>         ]
>       }
>     }
>   },
>   "_tooltips": {
>     "_isEnabled": true
>   },
> ```

Yup, need a solution for touch. If you wouldn't mind finding one. I was thinking one of the below but needs more research:
* off
* first tap show, second tap click
* long tap show

oliverfoster:
This should be fixed by the latest commit https://github.com/adaptlearning/adapt-contrib-core/pull/207/commits/63bd422c2401d4697a0a6c1e3dc3efe0dba3f4d6

Any focus assigned by a11y does not now trigger a tooltip. Such that, only focus assigned from a keyboard, mouseover, or click/tap, which retains focus for longer than 500ms, will show a tooltip. 

So focusable elements that only display a tooltip, and retain focus, will work everywhere (we have none of these yet, as all clicks move focus somehow, but it could be fun in future). Otherwise any use of a mouse or keyboard to hover or tab into focus, will show a tooltip, such that it's possible to have a touch device with a mouse and a keyboard and still get tooltips.

guywillis:
From reading a few sources:

* [material.io](https://m3.material.io/components/tooltips/guidelines#00e87770-86d0-436d-b50b-436ff3cefe75)
* [medium](https://medium.com/smarter-workforce-design/tooltips-in-the-era-of-touch-5268615bf321)
* [Adobe Spectrum](https://react-spectrum.adobe.com/react-spectrum/ContextualHelp.html)
* [Stack Exchange](https://ux.stackexchange.com/questions/98628/how-can-i-display-tool-tips-on-a-touchscreen-or-is-there-a-better-way)

The solutions mirror what you have laid out above - tap and hold, tap to show / tap to click, and off

Generally, it's advised to try not to have tooltips in a UI for desktop and mobile as the UI should explain itself without the need for further information. 

However, should tooltips be required the more suitable solution would be tap and hold over double clicking as it's more intuitive for end users - tap and hold is a fairly common input on touch devices as opposed to clicking the same button twice. The downside to tap and hold though is that it should only be applied to specific elements as to not interfere with browser default functionality of things like links and images. 

As per our offline conversation, we'll suggest going without tooltips for touch devices click events.

StuartNicholls:
When I looked at this, - so this is for standard tooltip functionality - I had intended to have tooltips all off, apart from on disabled buttons as that is a good accessibility feature and doesn't get it the way of normal clicks.

StuartNicholls:
> It's now on with all devices, on focus and mouseover, after the same 500ms
> 
> The bug has been fixed for having them now NOT show if focus is step by accessibility, such as, coming down from a popup.
> 
> They should only appear on a focus / mouseover after 500ms, when not interupted by a blur or mouseout. Which for most cases on mobile, will mean that they never show unless we've gone out of our way to retain focus on something clickable, that has a tooltip, for more than 500ms.

Yes, that was my understanding of how it worked when I worked on it. So I was talking about the outcome yep. maybe Im confused by this thread but I thought Guy was seeing tooltips on buttons on touch, I thought I'd seen them appearing in places we didn't necessarily want them also. I'll have another look tomorrow.

oliverfoster:
> I thought Guy was seeing tooltips on buttons on touch, I thought I'd seen them appearing in places we didn't necessarily want them also. I'll have another look tomorrow.

It's been on focus down from popups so far. There are no button with a tooltip that don't blur immediately when clicked, as they all navigate or open a popup across hotgraphic or navigation.

> If I **open and close the drawer** using the buttons provided (navigation drawer and drawer close buttons) I am shown the tooltip after closing.

> Similarly, if I select a hotspot within the hotgraphic interaction and **close the popup** using the subsequent notify close button the tooltip is present until I press the screen or interaction.

On android a long press on a button assigns it focus without a click and then the popup does show. But I'm happy with that, There isn't much I can do about it as longpress isn't a DOM event, it'd need a separate touch-event manager / longpress lib to overcome. On iOS a long press on a button does a text selection.

guywillis (approved):
👍

PR #213, opened 2022-08-03
Fix: Added an extra guard against a blank child in getAllDescendantModels

### LINKED ISSUE #212: Guard needed against empty child in adaptModel.getAllDescendantModels
We are seeing strange issues with large question banks that fall over when this._childrenCollection is returning arrays containing undefined

### Your environment
* FW5.22.0
* All browsers
* All devices

### Steps to reproduce
Difficult to say, but large assessments (40+ questions) that are using question banks. Somehow the this._childrenCollection is getting set to an array containing undefined. Difficult to debug exactly because of the sheer volume of loops over `setChildren()`

### Expected behaviour
Page should load as expected

### Actual behaviour
`adaptModel.js:471 Uncaught TypeError: Cannot read properties of undefined (reading 'hasManagedChildren') at adaptModel.js:471:18`

cahirodoherty-learningpool:

fixes #212 

### Fix
* Added an extra guard against a blank child in getAllDescendantModels
* Fixed release template typo

### Testing
1. Create a course with high volume of question components inside an assessment
2. Use question banks of varying sizes
3. Preview the course and ensure assessment loads as expected

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -468,6 +468,8 @@ export default class AdaptModel extends LockingModel {
 
     children.models.forEach(child => {
 
+      if (!child) return;
+
       if (!child.hasManagedChildren) {
         descendants.push(child);
         return;
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,7 @@
 * A sentence describing each fix
 
 ### Update
-* A sentence describing each udpate
+* A sentence describing each update
 
 ### New
 * A sentence describing each new feature


ethan-lp (approved):

js/models/adaptModel.js
```diff
@@ -468,6 +468,8 @@ export default class AdaptModel extends LockingModel {
 
     children.models.forEach(child => {
 
+      if (!child) return;
+
       if (!child.hasManagedChildren) {
```
eleanor-heath, line 473: Think we should add the optional chaining operator here instead of adding the new if conditional statement, e.g.`if (!child?.hasManagedChildren) {`

eleanor-heath (approved):

js/models/adaptModel.js
cahirodoherty-learningpool: This would still push an `undefined` onto the descendants array (after the if)

js/models/adaptModel.js
```diff
@@ -468,6 +468,8 @@ export default class AdaptModel extends LockingModel {
 
     children.models.forEach(child => {
```
oliverfoster, line 469: When is children.models getting populated with undefined? `[undefined]`

cahirodoherty-learningpool: Yeah it is concerning. I have not been able to pin it down while debugging. The sheer number of loops through `setChildren` (100's?) has made it difficult, and any conditional debugging (checking for undefined value for example) doesn't seem to kick in as I expect, so something strange going on.

I figured this would be a beneficial early return check that gets us past the problem of the course falling over

cahirodoherty-learningpool: @oliverfoster Any further thoughts on this?

oliverfoster: Sorry mate, I missed this comment.

PR #219, opened 2022-08-15
New: Drawer position, drawer API refactor (fixes #216, #217, #218)

### LINKED ISSUE #216: Drawer position independently from language direction
Sometime you need drawer to be on the left side during language direction "ltr". It will be really good to be able to change drawer position independently from language _defaultDirection.

maybe something like 
```
"_drawer":{
"_drawerPosition":"auto"/"left"/"right" / - "auto" -> to go with _defaultDirection 
} 
```

### LINKED ISSUE #217: Drawer View: add hooks
It would be quite useful when developing other plugins to be able to establish whether the drawer is open or not. E.g. before re-enabling scrolling.
At the moment there appear to be no hooks into drawerView so the above has to be determined by reading it's DOM attributes which is not ideal.

I'd suggest at least following the convention of other modules by triggering `drawerView:pre/postRender` events and maybe also passing in the view instance to the `drawer:opened` and `drawer:closed` triggers.

Maybe also move the _isVisible property to `Adapt.drawer` so the value is available globally.

Am happy to supply a PR for the above. Anyone have any further comments or requests before I do so?

### LINKED ISSUE #218: Prevent duplicate items being added to drawer
There's currently no way to know if an item already exists in a drawer (via the drawer interface). I think it would be good if the 'addItem' function on the drawer interface required an 'id' passed in it's data. A simple check would then ensure duplicates don't occur.

This identifier would also be useful to update or remove existing drawer items.

oliverfoster:
fixes #216 
fixes #217 
fixes #218 

### Fix
* Reworked main file into a proper controller with defined API

### New
* Added `_drawer._position = "auto|left|right"` to `config.json`
* Added `drawer.isOpen`
* Prevent duplicate items

### DIFF
--- a/schema/config.model.schema
+++ b/schema/config.model.schema
@@ -328,6 +328,20 @@
           "title": "Duration",
           "inputType": "Number",
           "validators": ["required", "number"]
+        },
+        "_position": {
+          type": "string",
+          "required": true,
+          "default": "auto",
+          "title": "Position",
+          "inputType": {
+            "type": "Select",
+            "options": [
+              "auto",
+              "left",
+              "right"
+            ]
+          }
         }
       }
     },
--- a/schema/config.model.schema
+++ b/schema/config.model.schema
@@ -312,7 +312,7 @@
           "validators": ["required", "number"]
         },
         "_position": {
-          type": "string",
+          "type": "string",
           "required": true,
           "default": "auto",
           "title": "Position",
--- a/templates/nav.hbs
+++ b/templates/nav.hbs
@@ -14,7 +14,7 @@
     <span class="icon"></span>
   </button>
 
-  <button class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn" data-event="toggleDrawer" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}">
+  <button class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{_config._drawer._position}}" data-event="toggleDrawer" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}">
     <span class="icon"></span>
   </button>
 
--- a/js/deprecated.js
+++ b/js/deprecated.js
@@ -45,6 +45,32 @@ Adapt.on({
 });
 /** END REDUNDANT ADAPT WAIT BEHAVIOUR */
 
+/** START REDUNDANT DRAWER BEHAVIOUR */
+Adapt.on({
+  'drawer:remove'() {
+    logging.log.deprecated('Use drawer.remove, Adapt.trigger(\'drawer:remove\') will be removed in the future');
+    drawer.remove();
+  },
+  'drawer:closeDrawer'() {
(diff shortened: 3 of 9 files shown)


templates/nav.hbs
```diff
@@ -14,7 +14,7 @@
     <span class="icon"></span>
   </button>
 
-  <button class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn" data-event="toggleDrawer" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}">
+  <button class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{_config._drawer._position}}" data-event="toggleDrawer" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}">
```
cahirodoherty-learningpool, line 17: Isn't this handled in the View className() already?

oliverfoster: No, that's for the child view of the drawer, this is the navigation button.

cahirodoherty-learningpool (approved):
👀

kirsty-hames (commented):
When `_drawer._position` is `left`, the drawer displays on the left regardless of language direction. 
When `_drawer._position `is `right`, the drawer only displays on the right for `ltr` language direction. I presume this should remain right and only `_drawer._position` `auto` should follow the text direction?

oliverfoster:
> When `_drawer._position` is `left`, the drawer displays on the left regardless of language direction. When `_drawer._position `is `right`, the drawer only displays on the right for `ltr` language direction. I presume this should remain right and only `_drawer._position` `auto` should follow the text direction?

This is true any suggetions?

schema/config.model.schema
```diff
@@ -328,6 +328,20 @@
           "title": "Duration",
           "inputType": "Number",
           "validators": ["required", "number"]
+        },
+        "_position": {
+          type": "string",
```
joe-allen-89, line 333: Missing opening quote on line 333.

joe-allen-89 (changes requested):
One slight mistype in schema but everything else working as expected. 

Question: When _defaultDirection is changed to "rtl", the navbar icons move to the left, should this also happen when _position is changed to "left"?

oliverfoster:
requestReview(@kirsty-hames && @joe-allen-89)

kirsty-hames:
> One slight mistype in schema but everything else working as expected.
> 
> Question: When _defaultDirection is changed to "rtl", the navbar icons move to the left, should this also happen when _position is changed to "left"?

I agree, I think most, if not all use cases you would want the drawer navbar icon to display on the same side the drawer opens. I think navbar icon positioning should be handled via navbar config though as this will also impact the ordering of nav bar icons. I think this will fall into https://github.com/adaptlearning/adapt-contrib-core/pull/239 PR.

kirsty-hames (approved):

joe-allen-89 (approved):
Tested in latest FW, agree with Kirsty's comment re the nav bar.

oliverfoster:
I don't think we should automate it. If you want the drawer on the left you'll need to set the button to appear on the left. The two things aren't related code-wise and I'm hesitant to make them so as it introduces more complexity in the `_navOrder` stuff. Once the two things go out, perhaps we can have another look?

kirsty-hames:
> 

I agree that the two should remain separate. My comment wasn't that clear but that's what I meant by the navbar icon positioning should be handled via navbar config.

PR #221, opened 2022-08-16
Update: Remove tabindex from a11y.toggleEnabled (fixes #220)

### LINKED ISSUE #220: Remove tabindex from A11y toggleEnabled()
When using a11y.toggleEnabled, a tabindex attribute of -1 is added.  This results in the button not being focusable via keyboard navigation.

This function also adds aria-disabled="false".  If we remove tabindex="-1" but keep aria-disabled="false", the user would be able to tab to the element and understand that it's disabled.  This may be a better user experience than not allowing the user to tab to the element altogether.

### Your environment
* Framework 5.x

### Steps to reproduce
Add a Hotgraphic component to a course.  Click on an item to open the Notify popup.  Note that the paging buttons call a11y.toggleAccessibleEnabled to enable or disable the buttons.  That function then calls a11y.toggleEnabled.

### Expected behaviour
Disabled buttons should be visible, focusable and marked as disabled.

### Actual behaviour
Disabled buttons are hidden and not focusable.

### Considerations for other plugins
As the disabled buttons would now be focusable, some plugin styles may need to be updated.  For instance, the Hotgraphic disabled paging buttons should probably be made visible (i.e. remove `visibility: hidden`) and styled to resemble a disabled button state.

swashbuck:
Fixes #220

### Update
* Removes tabindex from a11y.toggleEnabled

### Testing
1. Add a Hotgraphic component to a course.
2. Click on an item to open the Notify popup.
3. The disabled buttons should be focusable. Requires https://github.com/adaptlearning/adapt-contrib-hotgraphic/issues/241

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -300,11 +300,10 @@ class A11y extends Backbone.Controller {
     }
     if (!isEnabled) {
       $elements.attr({
-        tabindex: '-1',
         'aria-disabled': 'true'
       }).addClass('is-disabled');
     } else {
-      $elements.removeAttr('aria-disabled tabindex').removeClass('is-disabled');
+      $elements.removeAttr('aria-disabled').removeClass('is-disabled');
     }
     return this;
   }


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

kirsty-hames (approved):
👀

kirsty-hames:
On retest, I've found that this doesn't resolve the issue for Hotgraphic popup controls. The tabindex is still being rendered.

tabindex

The toggling is handled via [toggleAccessibleEnabled](https://github.com/adaptlearning/adapt-contrib-hotgraphic/blob/e02c954dfddb4705f1982a59547a55eb02b44a0f/js/hotgraphicPopupView.js#L47) which calls both `toggleEnabled` and `toggleAccessible`. I'm not entirely sure what the difference or use case for these are?

Narrative controls are handled via [toggleEnabled](https://github.com/adaptlearning/adapt-contrib-narrative/blob/248d59cc3b1a23ea0381a3391ed1471cc48df68d/js/NarrativeView.js#L273) - should we not be using the same for Hotgraphic for consistency @swashbuck / @oliverfoster ?

oliverfoster:
yes @kirsty-hames

PR #234, opened 2022-09-09
Fix: aria-level is set to "0" for headings (fixes #233)

### LINKED ISSUE #233: aria-level is set to "0" for headings
Since #146 has ariaLevel use ?? to check the level override, it considers the 0 default value to be a valid level, and ends up setting aria-level="0" on all headings.

hanshillen:
Fixes #233

### Fix

* Change nullish coalescing operator to logical OR for ariaLevel check, so that 0 is not accepted as a level value.

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -189,6 +189,8 @@ class A11y extends Backbone.Controller {
     }
     // get the global configuration from config.json
     const ariaLevels = Adapt.config.get('_accessibility')?._ariaLevels ?? defaultAriaLevels;
+    // Fix for authoring tool schema _ariaLevel = 0 default
+    if (override === 0) override = null;
     /**
      * Recursive function to calculate aria-level
      * @param {string} id Model id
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster:
I would rather warn that the heading level is invalid. Where is _ariaLevel beening set to 0? Is it human error or some default?

hanshillen:
It defaults to 0 in page.model.schema, article.model.schema, block.model.schema, and component.model.schema:

```json
    "_ariaLevel": {
      "type": "number",
      "default": 0,
      "isSetting": true,
      "inputType": "Number",
      "validators": ["number"],
      "help": "If you need to override the default component ARIA level (as defined in Configuration Settings), set this to any number greater than 0."
    },
```

So the AAT will set _ariaLevel to 0 for every model that it creates.

oliverfoster:
Hmm, can you just put an if block at the top of the function to correct this?
```js
if (override === 0) override = null;
```

hanshillen:
> Hmm, can you just put an if block at the top of the function to correct this?

Done

> The `_ariaLevel` type should be changed to a string really.

Agreed, and then check for an empty string instead

oliverfoster (approved):

StuartNicholls (approved):

PR #237, opened 2022-09-09
Breaking: Use heading elements (Fixes #235)

### LINKED ISSUE #235: Use heading elements
Headings are currently marked up as follows:

```html
<div class="js-heading is-incomplete">
    <div id="c-15-heading" class="js-heading-inner" role="heading" aria-level="0">
        <span class="aria-label">Incomplete. My title</span>
    </div>
</div>
<div class="component__title-inner text__title-inner" aria-hidden="true">My title</div>
```
So the actual heading is visually hidden, while the visible heading is programmatically hidden from assistive tech. This in itself is not  semantically incorrect, but the approach has a few side effects:

* Any assistive tech product, tool, extension, or user-defined style sheet that targets headings (by highlighting them or overriding their styles) won't be able to do so; the programmatically detectable heading is not visible (and its bounds do not match the bounds of the visible heading).
* Since the visible heading is the same selector regardless of its level, there's no easy selector available to style heading levels individually (e.g. to make an h2 bigger than an h3). Using classes like `component__title-inner` won't work, since that heading is not guaranteed to be the same level for every component. The `aria-level` attribute is on the invisible element, so not suitable for a selector either (unless you do something crazy like `.js-heading:has(>[aria-level="2"])+.component__title-inner)` which won't work anyway).
* It's unnecessarily complicated and redundant, without offering any benefits over regular headings.

In my theme I've changed the templates to use heading elements instead (using the ariaLevel value for the element name). Making them look like this:

```html
<div class="js-heading component__title-inner text__title-inner is-incomplete">
  <h3 id="c15-heading" class="js-heading-inner">     
    <span class="aria-label">
        Incomplete.
    </span>
    My title
  </h3>
</div>
```

I think it would be useful to use this approach in core as well, as it resolves the side effects listed above while preserving the semantics. 

Note: This is a feature request. not a bug. If the change is not feasible because it would break other plugins (like themes and pageLevelProgress), then this issue can be closed. The issue and associated PR can still be useful for others looking to do the same in their theme in the future.

hanshillen:
Fixes #235

### Update

* Change old heading markup with heading elements.

### Breaking

* This will break other plugins that depend on the old heading structure (e.g. 
themes, menus, pageLevelProgress).

### DIFF
--- a/templates/article.hbs
+++ b/templates/article.hbs
@@ -8,16 +8,8 @@
     <div class="article__header-inner">
 
       {{#if displayTitle}}
-      <div class="article__title">
-
-        {{#unless _disableAccessibilityState}}
-        <div class="js-heading"></div>
-        {{~/unless~}}
-
-        <div class="article__title-inner"{{#unless _disableAccessibilityState}} aria-hidden="true"{{/unless}}>
-          {{{compile displayTitle}}}
-        </div>
-
+      <div class="article__title js-heading-container">
+        <div class="js-heading article__title-inner"></div>
       </div>
       {{/if}}
 
--- a/templates/block.hbs
+++ b/templates/block.hbs
@@ -8,16 +8,8 @@
     <div class="block__header-inner">
 
       {{#if displayTitle}}
-      <div class="block__title">
-
-        {{#unless _disableAccessibilityState}}
-        <div class="js-heading"></div>
-        {{~/unless~}}
-
-        <div class="block__title-inner"{{#unless _disableAccessibilityState}} aria-hidden="true"{{/unless}}>
-          {{{compile displayTitle}}}
-        </div>
-
+      <div class="block__title js-heading-container">
+        <div class="js-heading block__title-inner"></div>
       </div>
       {{/if}}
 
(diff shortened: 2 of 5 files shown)


oliverfoster:
Can you make it not breaking?

hanshillen:
> Can you make it not breaking? The primary goal doesn't require you to break anything.

I'm not sure it can be done solely within core.  The BoxMenu plugin would still have the old template for menu and menu item headings in boxMenuItem.hbs. That template would need to remove the aria-hidden displayTitle.  

> Why do you want to switch to symantic heading tags? They come with browser styling that needs to be overridden immediately

I've provided a detailed rationale in #235 

> but accessibility wise, they behave identically to role=heading aria-level=n, 

I would argue the same as a reason for this change (per the [first rule of ARIA](https://w3c.github.io/using-aria/#rule1)): The same semantics can be achieved with a much less convoluted approach, in a way that is backward compatible to the easily days of the web, easy to understand and straightforward to style by course designers, plugin developers, students, and AT tools / browser extensions alike. As I mentioned in #235, there's currently no clear method to style different heading levels separately by any of these actors (or have them pointed out at all.

> it's just that headings go from 1-6 whereas aria-level is unbounded.

The lack of heading level bounds does not provide any benefits. Any document environment (e.g. PDF, Word, HTML) uses six heading levels for accessibility. Screen readers are set up to work with these six levels. For example, JAWS maps <kbd>1</kbd> to <kbd>6</kbd> as shortcuts to jump directly to the next heading of that level. It does not have a skipping mechanism for a level 7 heading, and will in fact wrap around and treat that heading as level 2. In general, a heading structure with a >6 nesting is indicative of a structural issue that would have to be resolved.

oliverfoster:
Merging into separate branch to work on.

PR #239, opened 2022-09-14
Fix: Navigation button read order (fixes #238)

### LINKED ISSUE #238: Navigation bar buttons accessibility
### Subject of the issue
Need to switch from `float` styling to something that retains the dom order more effectively. 
Dom order can currently be set with a `data-order` attribute on each button but the styling breaks the visual order.

Reference https://github.com/adaptlearning/adapt-contrib-core/pull/61

oliverfoster:
fixes #238
requires https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/151
requires https://github.com/adaptlearning/adapt_framework/pull/3332

The dom order and visual order will be aligned, such that the screen reader read order is correct.

### Fix
* Moved from `float` styling to `flex`
* Added spacers
* Added default schema order for buttons and spacers
* Reorder happens when any child is changed

### DIFF
--- a/js/views/navigationView.js
+++ b/js/views/navigationView.js
@@ -3,6 +3,7 @@ import a11y from 'core/js/a11y';
 import location from 'core/js/location';
 import router from 'core/js/router';
 import startController from 'core/js/startController';
+import _ from 'underscore';
 
 class NavigationView extends Backbone.View {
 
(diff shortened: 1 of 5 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👍

kirsty-hames (dismissed):

kirsty-hames (commented):
Just to expand on my previous comment, changing `_navOrder` in the _course.json_ doesn't seem to do anything and instead the schema 'default' is being applied. This issue doesn't apply to the [PLP PR ](https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/151)however, this updates as expected via `_navOrder`.

StuartNicholls (changes requested):
[schema/course.model.schema] line 337. 

When compiled the _extensions object incorrectly sits within the _accessibility object. Should be at same level as _accessibility in the JSON

oliverfoster:
@kirsty-hames I'd put the `_extensions` object in the wrong place in `src/course/en/course.json`, it should have been under `_accessibility` not in it, as @StuartNicholls has detailed above

PR #247, opened 2022-10-05
Fix: Made headings clickable

### LINKED ISSUE #235: Use heading elements

oliverfoster:
fixes #235 

Instead of using semantic heading levels, as this requires a lot of styling fixes in vanilla, I've focused more on making the headings clickable and unified, rather than having a separate heading for accessibility.

### Fix
* Unify accessible and display headings

Requires 
https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/152
https://github.com/adaptlearning/adapt-contrib-boxMenu/pull/127

### DIFF
--- a/templates/heading.hbs
+++ b/templates/heading.hbs
@@ -1,6 +1,6 @@
 {{import_globals}}
 
-<div id="{{_id}}-heading" role="heading" aria-level="{{a11y_aria_level _id _type _ariaLevel}}">
+<div id="{{_id}}-heading" role="heading" class="js-heading-inner" aria-level="{{a11y_aria_level _id _type _ariaLevel}}">
   <span class="aria-label">
   {{#if _isA11yCompletionDescriptionEnabled}}
     {{#if _isOptional}}
--- a/templates/heading.hbs
+++ b/templates/heading.hbs
@@ -1,6 +1,6 @@
 {{import_globals}}
 
-<div id="{{_id}}-heading" role="heading" class="js-heading-inner" aria-level="{{a11y_aria_level _id _type _ariaLevel}}">
+<div id="{{_id}}-heading" class="js-heading-inner" role="heading" aria-level="{{a11y_aria_level _id _type _ariaLevel}}">
   <span class="aria-label">
   {{#if _isA11yCompletionDescriptionEnabled}}
     {{#if _isOptional}}
--- a/templates/heading.hbs
+++ b/templates/heading.hbs
@@ -1,19 +1,16 @@
 {{import_globals}}
 
-<div id="{{_id}}-heading" class="js-heading-inner" role="heading" aria-level="{{a11y_aria_level _id _type _ariaLevel}}">
-
+<h{{a11y_aria_level _id _type _ariaLevel}} id="{{_id}}-heading" class="js-heading-inner">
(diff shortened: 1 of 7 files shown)


templates/heading.hbs
```diff
@@ -1,19 +1,21 @@
...
 
 <div id="{{_id}}-heading" class="js-heading-inner" role="heading" aria-level="{{a11y_aria_level _id _type _ariaLevel}}">
-
   <span class="aria-label">
   {{#if _isA11yCompletionDescriptionEnabled}}
     {{#if _isOptional}}
-      {{{compile title}}}
       {{else if _isComplete}}
-      {{_globals._accessibility._ariaLabels.complete}}. {{{compile title}}}
+      {{_globals._accessibility._ariaLabels.complete}}.&nbsp;
```
cahirodoherty-learningpool, line 8: Is this &nbsp; coming from your editor?

oliverfoster: It's there because a space has to exist between the `Complete.` and `{{compile displayTitle}}`. Otherwise NVDA will read the period aloud, "CompleteDOTTitle" rather than "Complete[Pause]Title".

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
Tested with the latest version of FW, Chrome and NVDA, working as expected, a very nice change for accessibility.

PR #252, opened 2022-10-19
Fix: Corrected check for jsx template (fixes #224)

### LINKED ISSUE #224: Incorrect jsx template check
Should be:

if (this.isJSX) {

https://github.com/adaptlearning/adapt-contrib-core/blob/3790f356f1a65f93f880a81db8b21ab29f2de07d/js/views/contentObjectView.js#L40

### Your environment
Latest version

### Steps to reproduce
Doesn't work with jsx menus

oliverfoster:
fixes #224 

### Fix
* Corrected check for jsx template

### DIFF
--- a/js/views/contentObjectView.js
+++ b/js/views/contentObjectView.js
@@ -37,7 +37,7 @@ export default class ContentObjectView extends AdaptView {
     const type = this.constructor.type;
     Adapt.trigger(`${type}View:preRender contentObjectView:preRender view:preRender`, this);
 
-    if (this.isReact) {
+    if (this.isJSX) {
       this.changed();
     } else {
       const data = this.model.toJSON();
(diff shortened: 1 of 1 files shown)


joe-allen-89 (approved):

js/views/contentObjectView.js
```diff
@@ -37,7 +37,7 @@ export default class ContentObjectView extends AdaptView {
     const type = this.constructor.type;
     Adapt.trigger(`${type}View:preRender contentObjectView:preRender view:preRender`, this);
 
-    if (this.isReact) {
+    if (this.isJSX) {
```
StuartNicholls, line 40: @oliverfoster Do we need to make this change to line 150 also?

PR #255, opened 2022-10-31
Fix: Reset _byAdaptID on load manifest files (fixes #253)

### LINKED ISSUE #253: loadManifestFiles should empty _byAdaptID to avoid stale references
the `_byAdaptID` object may contain stale references immediately after language change. Note that although `Backbone.Collection.reset` is called, `remove` events are [not fired](https://backbonejs.org/#Collection-reset) and so `_byAdaptID` references will not be removed by `data.onRemoved`.

One example of stale references occuring is with [adapt-contrib-branching](https://github.com/adaptlearning/adapt-contrib-branching) on language change. This plugin waits for all `app:dataReady` handlers to execute so potentially these handlers will see branching models created for the previously selected language.

### Your environment
* all recent framework versions
* all devices and browsers

### Expected behaviour
`loadManifestFiles` should empty `_byAdaptID`.

### Actual behaviour
See description

oliverfoster:
fixes #253 

### Fix
* Reset _byAdaptID on load manifest files

### DIFF
--- a/js/data.js
+++ b/js/data.js
@@ -120,6 +120,7 @@ class Data extends AdaptCollection {
   async loadManifestFiles(languagePath) {
     this.trigger('loading');
     this.reset();
+    this._byAdaptID = {};
     const manifestPath = languagePath + 'language_data_manifest.js';
     let manifest;
     try {


cahirodoherty-learningpool (approved):
👀

chris-steele (approved):

joe-allen-89 (approved):

PR #256, opened 2022-10-31
Fix: Allow locking on object specified attributes, performance enhancement (fixes #183)

### LINKED ISSUE #183: set locked attribute loophole
A locked attribute can be overriden if it is set within an object as this doesn't pass through the locking checks:
https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/models/lockingModel.js#L4

oliverfoster:
fixes #183 

### Fix
* Do not attempt to update and trigger attributes which haven't changed
* Update attributes in batches rather than individually
* Allow locking attributes to be changed via object syntax


cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

js/models/lockingModel.js
```diff
@@ -1,33 +1,55 @@
...
+      }
+
+      const isSettingValueForSpecificPlugin = options?.pluginName;
+      if (!isSettingValueForSpecificPlugin) {
+        console.error('Must supply a pluginName to change a locked attribute');
+        options.pluginName = 'compatibility';
+      }
+
+      const pluginName = options.pluginName;
+      if (this.defaults[attrName] !== undefined) {
```
chris-steele, line 29: I feel like I have missed something, but `defaults` are generally provided via a function so presumably array notation will not work as intended?

chris-steele (commented):
Looking at [this line](https://github.com/adaptlearning/adapt-contrib-core/pull/256/files#diff-828c154036fceb2feb0ed7365f50e3a968b9b30923ce48e77993eec7a3d1fccdL55), where is `this.lockedAttributes` defined?

js/models/lockingModel.js
oliverfoster: I agree. Will change. Ta.

oliverfoster:
> Looking at [this line](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/models/lockingModel.js#L55), where is `this.lockedAttributes` defined?

https://github.com/adaptlearning/adapt-contrib-core/search?q=lockedAttributes

eleanor-heath (approved):

PR #258, opened 2022-10-31
Fix: Allow plugins to pause rendering (fixes #257)

### LINKED ISSUE #257: JSX questions cause multiple unnecessary render events on submit
### Subject of the issue
Due to the nature of the `onSubmitClicked` function, jsx templates will attempt to rerender on every property change during this process.
https://github.com/adaptlearning/adapt-contrib-core/blob/5ef489512f19984a40a158d69b23a7fe410f7e7d/js/views/questionView.js#L181-L229

oliverfoster:
fixes #257 

### Fix
* Allow plugins to pause rendering
* Stop and start rendering in `onSubmitClicked` handler to reduce cpu usage

### DIFF
--- a/js/views/adaptView.js
+++ b/js/views/adaptView.js
@@ -18,6 +18,7 @@ class AdaptView extends Backbone.View {
   }
 
   initialize() {
+    this._jsxIgnoreChanges = 0;
     this.listenTo(this.model, {
       'change:_isVisible': this.toggleVisibility,
       'change:_isHidden': this.toggleHidden,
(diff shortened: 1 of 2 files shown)


joe-allen-89 (approved):

ethan-lp (approved):

chris-steele:
Just a suggestion, but would it be simpler to just debounce `changed`?

oliverfoster:
The view should update synchronously on model change, debounce would make it asynchronous.

PR #273, opened 2022-11-24
New: Allow plugins to set the drawer position

### LINKED ISSUE #272: Allow plugins to specify drawer position
In addition to the changes made https://github.com/adaptlearning/adapt-contrib-core/pull/219, plugins may wish to set their own drawer position when creating a custom view.

### Proposed changes
The global `_drawer._position` is applied to all views opened in the drawer, as is the case currently:

- `"auto"` will switch positions accordingly based on left-to-right/right-to-left language directions

- `"left"` or `"right"` will always be positioned to match the setting regardless of language direction

Additionally, a plugin can set the position for its own view via `openCustomView`. As `_drawer._position` is a global setrting for all drawer views, the position will only be inverted for right-to-left languages if `_drawer._position: "auto"`.

danielghost:
Fixes #272

### New
* Allow plugins to set the drawer position for a custom view.

### DIFF
--- a/js/views/drawerView.js
+++ b/js/views/drawerView.js
@@ -64,22 +64,19 @@ class DrawerView extends Backbone.View {
   }
 
   setDrawerPosition(position) {
+    const isGlobalPositionAuto = this._globalDrawerPosition === 'auto';
+    const isRTL = Adapt.config.get('_defaultDirection') === 'rtl';
+    if (position && isGlobalPositionAuto && isRTL) position = (position === 'left') ? 'right' : 'left';
+    if (!isGlobalPositionAuto || !position) position = this._globalDrawerPosition;
     this.$el
       .removeClass(`is-position-${this.drawerPosition}`)
       .addClass(`is-position-${position}`);
     this.drawerPosition = position;
-    this.drawerAnimationDir = (position === 'auto') ? ((Adapt.config.get('_defaultDirection') === 'rtl') ? 'left' : 'right') : position;
-  }
-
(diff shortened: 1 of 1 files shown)


js/views/drawerView.js
```diff
@@ -72,8 +63,24 @@ class DrawerView extends Backbone.View {
...
-  openCustomView(view, hasBackButton) {
+  setDrawerPosition(position) {
+    this.$el
+      .removeClass(`is-position-${this.drawerPosition}`)
+      .addClass(`is-position-${position}`);
+    this.drawerPosition = position;
+    this.drawerAnimationDir = (position === 'auto') ? ((Adapt.config.get('_defaultDirection') === 'rtl') ? 'left' : 'right') : position;
+  }
+
+  getInversePosition(position) {
```
oliverfoster, line 74: This function looks like an afterthought.

Could you move all the position stuff out out openCustomView, change setDrawerPosition to setUpDrawerPosition and either move the code body of getInversePosition into setUpDrawerPosition or rename `getInversePosition` to `get position()` including the `_globalDrawerPosition` behaviour inside it?

danielghost: Have moved all the positional code into `setDrawerPosition` to remove `getInversePosition`. Haven't used a getter for position as we need to determine if a position parameter has been set by a plugin.

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

oliverfoster (approved):

PR #274, opened 2022-12-01
Update: Updated focus to focus-visible (fixes #231)

### LINKED ISSUE #231: Look at focus-visible
### Subject of the issue
Rather than relying on [keyboard detection](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/a11y/keyboardFocusOutline.js), use https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible instead.

Will require changes to all `:focus` selectors.

Plugins that require updating
- [ ] Core
- [ ] Media
- [x] Narrative
- [x] Slider
- [ ] Vanilla

guywillis:
Fixes: #231 

### Update
* Converted `:focus` to `:focus-visible`

### DIFF
--- a/less/core/nav.less
+++ b/less/core/nav.less
@@ -25,7 +25,7 @@
       .u-display-none;
     }
 
-    &:not(:focus) {
+    &:not(:focus-visible) {
       height: 0;
       padding: 0;
       margin: 0;


joe-allen-89 (commented):
There's another :focus in accessibility.less which may also need changing.

StuartNicholls (approved):

cahirodoherty-learningpool (approved):
👀

PR #280, opened 2022-12-15
Fix: `_isOptional` completion

### LINKED ISSUE #279: `_isOptional` completion
If a parent model is set to _isOptional AND its immediate children are also set to be _isOptional, simply launching a model with the same parent will mark the other optional models as completed.

This issue is dependant on the course structure and is less likely since https://github.com/adaptlearning/adapt_framework/commit/61ebb6bfdbd5a0a1c2670f099e0a1be4ed33378e, but it is still very possible.

danielghost:
Fixes #279.

### Fix
* Account for every child being `_isOptional` when checking/setting completion.

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -312,19 +312,20 @@ export default class AdaptModel extends LockingModel {
   }
 
   /**
-   * Function for checking whether the supplied completion attribute should be set to true or false.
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

chris-steele (approved):

joe-allen-89 (approved):

PR #281, opened 2022-12-16
New: ChildView API events

### LINKED ISSUE #259: `view:childAdded` of type trigger
Often we only need to listen to `view:childAdded` events for specific type groups, so it may be worth considering triggering an additional [event](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/views/adaptView.js#L211) with the model `_type` appended in the event name, so we don't have to return early from the generic event for views which may not be in scope?

danielghost:
Fixes #259.

### New
* Added model `_type` to the `view:addChild` and `view:childAdded` events, so listeners can be setup for specific model types.

### DIFF
--- a/js/views/adaptView.js
+++ b/js/views/adaptView.js
@@ -193,11 +193,13 @@ class AdaptView extends Backbone.View {
    * @returns {AdaptView} Returns this childView
    */
   addChildView(childView) {
+    const childModel = childView.model;
+    const type = childModel.get('_type')
     const childViews = this.getChildViews() || [];
     childViews.push(childView);
     this.setChildViews(childViews);
     const $parentContainer = this.$(this.constructor.childContainer);
-    switch (childView.model.get('_renderPosition')) {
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

chris-steele (commented):
With this change we'll have typed events for everything except `view:childAdded` and `view:requestChild`. For consistency should we provide typed events for these also?

oliverfoster:
@danielghost see @chris-steele comment about `view:requestChild`

danielghost:
@chris-steele - `view:requestChild` is only called when a model is not supplied to `_getAddChildEvent`. This event is used for things like branching which determine the next model to add, so I excluded it as the `_type` would not be defined.

oliverfoster (approved):

chris-steele:
@danielghost makes sense thanks 👍

chris-steele (approved):

PR #283, opened 2023-01-06
Fix: restore responses for incomplete questions

### LINKED ISSUE #282: restoring attempts and responses for incomplete questions
For spoor tracked courses, question data is currently only saved on `'change:_isInteractionComplete'`. This means `_attemptsLeft` is not incremented, so in theory you could have unlimited attempts at a question if you kept closing the content after incorrectly answering a question, assuming it was set to have more than 1 attempt initially.

If the content is setup to restore responses, this is only applied to questions where `_isInteractionComplete: true`, otherwise the question is reset.

Probably an edge case in terms of someone actually doing this, but should be preventable.

### Steps to reproduce
- Questions must have `"_attempts": 2` or higher.

- If using an assessment, it must be set to `"_isResetOnRevisit": false`.

- Tracking plugin must be set to `"_shouldStoreResponses": true`.

Browser sessions
- Answer a question incorrectly, don't reset; close the content and relaunch, returning to the question. You will still have 2 attempts at the question. In the case of tracking plugins which do store responses before the question is completed, the `_userAnswer` will have been reset, so the responses are not restored.

Page sessions
- If tested within the same session, answer a question incorrectly, don't reset; leave the page and revisit, returning to the question. `_attemptsLeft` will be restored, but the user responses will be reset.

### Expected behaviour
Assuming the responses have been correctly saved across sessions, the following should probably use `_isSubmitted` rather than `_isInteractionsComplete` to prevent the question from being reset:
https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/questionView.js#L85

danielghost:
Fixes #282.

### DIFF
--- a/js/views/questionView.js
+++ b/js/views/questionView.js
@@ -107,8 +107,7 @@ class QuestionView extends ComponentView {
    * @private
    */
   ensureLegacyLifecycleState() {
-    const isInteractionComplete = this.model.get('_isInteractionComplete');
-    if (isInteractionComplete) {
+    if (this.model.get('_isSubmitted')) {
       this.onHideCorrectAnswerClicked();
       return;
     }


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

eleanor-heath (approved):

PR #291, opened 2023-01-18
Fix: Fallback to incorrect values when partlyCorrect is empty

### LINKED ISSUE #290: Partly correct feedback fallback values needed
Partly correct feedback should fall back to incorrect feedback values

### Your environment
* FW v5.24.4

### Steps to reproduce
Create a course with MCQ question with multiple correct answers.
Leave partly correct feedback empty
Preview course and answer partially correct/incorrect

### Expected behaviour
Incorrect feedback should be displayed

### Actual behaviour
No feedback displayed

cahirodoherty-learningpool:

Fixes #290 

### Fix
* Added fallback values to partlyCorrect feedback. Appropriate incorrect feedback is displayed

### DIFF
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -194,7 +194,6 @@ class QuestionModel extends ComponentModel {
   // Checks if the question should be set to complete
   // Calls setCompletionStatus and adds complete classes
   checkQuestionCompletion() {
-
     const isComplete = (this.get('_isCorrect') || this.get('_attemptsLeft') === 0);
 
     if (isComplete) {
@@ -202,13 +201,11 @@ class QuestionModel extends ComponentModel {
     }
 
     return isComplete;
-
   }
 
   // Updates buttons based upon question state by setting
   // _buttonState on the model which buttonsView listens to
   updateButtons() {
-
     const isInteractionComplete = this.get('_isInteractionComplete');
     const isCorrect = this.get('_isCorrect');
     const isEnabled = this.get('_isEnabled');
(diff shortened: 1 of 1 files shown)


ethan-lp (approved):

eleanor-heath (approved):

js/models/questionModel.js
```diff
@@ -297,12 +294,12 @@ class QuestionModel extends ComponentModel {
             title,
             _classes,
             body: !isFinal
-              ? _feedback._partlyCorrect.notFinal
-              : _feedback._partlyCorrect.final
+              ? _feedback._partlyCorrect.notFinal || _feedback._incorrect?.notFinal || ''
+              : _feedback._partlyCorrect.final || _feedback._incorrect?.final || ''
```
eleanor-heath, line 298: ? _feedback._partlyCorrect?.notFinal || _feedback._incorrect?.notFinal || ''
: _feedback._partlyCorrect?.final || _feedback._incorrect?.final || ''

eleanor-heath (approved):

js/models/questionModel.js
cahirodoherty-learningpool: Is there a difference in your suggestion to what is already in there? I can't see it but could easily be missing it 😂

eleanor-heath: yes the suggestion of by adding the optional chaining operator with adding the extra `?` beside _feedback._partlyCorrect. In the same way you have added the operator beside the `_feedback._incorrect` object. It ensures that the _feedback._partlyCorrect object exists before trying to retrieve the value of the `notFinal` or `final` property.  If the object accessed or function called is undefined or null, it returns undefined instead of throwing an error. I added this suggestion to avoid an error being thrown should _feedback._partlyCorrect be null or undefined.

I have also noticed the following code in the `questionModel`:
`case 'partlyCorrect': {
        if (typeof _feedback._partlyCorrect === 'object') {
          // old style
          return {
            // add higher values
            altTitle,
            title,
            _classes,
            body: !isFinal
              ? _feedback._partlyCorrect.notFinal
              : _feedback._partlyCorrect.final
          };
        }
        // new style
        const feedbackPartlyCorrect = !isFinal ? _feedback._partlyCorrectNotFinal : _feedback._partlyCorrectFinal;`

this is open to debate, but we might want to also consider adding `_feedback._partlyCorrectNotFinal`  and `_feedback._partlyCorrectFinal` into the equation so doing something like this:

`? _feedback._partlyCorrect?.notFinal || _feedback._partlyCorrectNotFinal || _feedback._incorrect?.notFinal || ''
              : _feedback._partlyCorrect?.final || _feedback._partlyCorrectFinal` || _feedback._incorrect?.final || ''`

cahirodoherty-learningpool: Added that chaining operator. Good catch 👍

oliverfoster (approved):

PR #294, opened 2023-01-30
Fix: Enable Show Feedback button when feedback content only has a title (fixes #265)

### LINKED ISSUE #265: Show Feedback button is disabled when _feedback lacks body content
If _feedback is configured to only have title text and no body text, the "Show Feedback" button is disabled.

### Your environment
* Framework 5.24.8

### Steps to reproduce
Set up feedback for a question as follows:

```
"_feedback": {
  "altTitle": "",
  "title": "Feedback global title",
  "_classes": "",
  "_correct": {
    "altTitle": "",
    "title": "Correct feedback title",
    "body": ""
  },
  "_incorrectNotFinal": {
    "altTitle": "",
    "title": "",
    "body": "Incorrect feedback body, not final"
  },
  "_incorrectFinal": {
    "altTitle": "",
    "title": "",
    "body": "Incorrect feedback body, final"
  }
},
```

### Expected behaviour
"Show Feedback" button should open the Notify popup when clicked.

### Actual behaviour
"Show Feedback" button is disabled.

swashbuck:
Fixes #265

### Fix
* Add another event listener for when `feedbackTitle` changes. This allows the feedback button to be enabled when the feedback content only has a title and no body text.
* `onFeedbackMessageChanged()` is now only concerned with _enabling_ (not disabling) the Feedback button.
* In `refresh()`, use the new method `disableFeedbackButton()` instead of `onFeedbackMessageChanged()`.

### Testing
1. Include the following feedback in a question:

```
"_feedback": {
  "altTitle": "",
  "title": "Feedback global title",
  "_classes": "",
  "_correct": {
    "altTitle": "",
    "title": "Correct feedback title",
    "body": ""
  },
  "_incorrectNotFinal": {
    "altTitle": "",
    "title": "",
    "body": "Incorrect feedback body, not final"
  },
  "_incorrectFinal": {
    "altTitle": "",
    "title": "",
    "body": "Incorrect feedback body, final"
  }
},
```
2. Set `_canShowFeedback` to `true`
3. Ensure the Show Feedback button can be clicked when either answering correctly (title text only) or incorrectly (body text only).

### DIFF
--- a/js/views/buttonsView.js
+++ b/js/views/buttonsView.js
@@ -198,7 +198,7 @@ export default class ButtonsView extends Backbone.View {
     this.checkResetSubmittedState();
     this.checkFeedbackState();
     this.onButtonStateChanged(null, this.model.get('_buttonState'));
-    this.disableFeedbackButton();
+    this.onFeedbackMessageChanged(null, this.model.get('_isSubmitted'));
   }
 
 }
--- a/js/views/buttonsView.js
+++ b/js/views/buttonsView.js
@@ -86,9 +86,8 @@ export default class ButtonsView extends Backbone.View {
   onFeedbackMessageChanged(model, changedAttribute) {
     if (!this.model.get('_canShowFeedback')) return;
 
-    if (changedAttribute) {
-      this.enableFeedbackButton();
-    }
+    if (!changedAttribute) return;
+    this.enableFeedbackButton();
   }
 
   enableFeedbackButton() {
--- a/js/views/buttonsView.js
+++ b/js/views/buttonsView.js
@@ -1,7 +1,7 @@
 import Adapt from 'core/js/adapt';
 import a11y from 'core/js/a11y';
 import BUTTON_STATE from 'core/js/enums/buttonStateEnum';
-import logging from 'core/js/logging'
+import logging from 'core/js/logging';
 
 // convert BUTTON_STATE to property name
 const textPropertyName = {
@@ -84,13 +84,14 @@ export default class ButtonsView extends Backbone.View {
   }
 
   onFeedbackMessageChanged(model, changedAttribute) {
-    if (!this.model.get('_canShowFeedback')) return;
-
     if (!changedAttribute) return;
+
     this.enableFeedbackButton();
   }
 
   enableFeedbackButton() {
+    if (!this.model.get('_canShowFeedback')) return;
+
     a11y.toggleEnabled(this.$('.js-btn-feedback'), true);
   }
 
--- a/js/views/buttonsView.js
+++ b/js/views/buttonsView.js
@@ -22,6 +22,7 @@ export default class ButtonsView extends Backbone.View {
     this.listenTo(this.model, {
       'change:_buttonState': this.onButtonStateChanged,
       'change:feedbackMessage': this.onFeedbackMessageChanged,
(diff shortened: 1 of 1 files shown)


js/views/buttonsView.js
```diff
@@ -191,7 +198,7 @@ export default class ButtonsView extends Backbone.View {
     this.checkResetSubmittedState();
     this.checkFeedbackState();
     this.onButtonStateChanged(null, this.model.get('_buttonState'));
-    this.onFeedbackMessageChanged(null, this.model.get('feedbackMessage'));
+    this.disableFeedbackButton();
```
cahirodoherty-learningpool, line 201: I think there are scenarios where this feedback could be enabled.
The extra feedbackTitle changed option does not need to be considered here as the passed attribute is forcing an enable, so this line of code can remain how it was, or possibly updated to:
`this.onFeedbackMessageChanged(null, true);`

```toggleFeedbackButton(enabled) {
    a11y.toggleEnabled(this.$('.js-btn-feedback'), enabled);
  }```

Comment (swashbuck) [reply js/views/buttonsView.js]: Thanks, @cahirodoherty-learningpool !

So, on the master branch, "Show Feedback" is disabled when you first view the question. Once the question is answered or the attempts are used up, then "Show Feedback" is enabled (aside from the #265  issue).

> I think this refresh should always be forcing feedback as enabled. That seems to be the current behaviour anyway.

I don't believe we would always want the "Show Feedback" button to be enabled. Instead, what about enabling it after the question has been submitted? I modified your suggestion to include this, and it seems to work. There may be a specific use case that I'm not considering, though, so let me know what you think.

`this.onFeedbackMessageChanged(null, this.model.get('_isSubmitted'));`

**Review by cahirodoherty-learningpool — APPROVED** [review]
👀

File: js/views/buttonsView.js [hunk 2]
```diff
@@ -83,12 +84,18 @@ export default class ButtonsView extends Backbone.View {
...
   onFeedbackMessageChanged(model, changedAttribute) {
-    if (changedAttribute && this.model.get('_canShowFeedback')) {
-      // enable feedback button
-      a11y.toggleEnabled(this.$('.js-btn-feedback'), true);
-      return;
+    if (!this.model.get('_canShowFeedback')) return;
+
+    if (changedAttribute) {
+      this.enableFeedbackButton();
     }
```
oliverfoster, line 91: This is the main action of the function, it shouldn't be nested in an if block.
Use return early.
```

```suggestion
    if (!changedAttribute) return;
    this.enableFeedbackButton();
```

js/views/buttonsView.js
```diff
@@ -83,12 +84,18 @@ export default class ButtonsView extends Backbone.View {
   }
 
   onFeedbackMessageChanged(model, changedAttribute) {
-    if (changedAttribute && this.model.get('_canShowFeedback')) {
-      // enable feedback button
-      a11y.toggleEnabled(this.$('.js-btn-feedback'), true);
-      return;
+    if (!this.model.get('_canShowFeedback')) return;
```
oliverfoster, line 87: Remove this line from `onFeedbackMessageChanged` and put it at the top of `enableFeedbackButton`

oliverfoster (changes requested):

oliverfoster (approved):

eleanor-heath (approved):

PR #304, opened 2023-02-20
Fix: _isOptional parents with _isOptional children (fixes #303)

### LINKED ISSUE #303: _isOptional/completion bug
### Subject of the issue

In this [pr](https://github.com/adaptlearning/adapt-contrib-core/commit/a1d078a350adfd6cae7d3ff5801c9d29d567745b), code was added to make sure that optional pages work correctly, however it seems to result in forcing optional component to become mandatory in all types of parent, when the parent only has optional children. 

Suggestion: I think it should only do that if the parent is also optional.

line 327

i.e. on a mandatory block with optional, interactive children, you'll see that the block, article and page won't complete unless interactions are complete.

https://github.com/adaptlearning/adapt-contrib-core/blob/a1d078a350adfd6cae7d3ff5801c9d29d567745b/js/models/adaptModel.js#L320-L340

oliverfoster:
fixes #303 

When a parent is optional, its completion is only used for plp, wait for the children to complete before completing the parent (i.e. force them to become mandatory).

### Fix
* Only _isOptional parents will now force the children to become mandatory

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -325,8 +325,8 @@ export default class AdaptModel extends LockingModel {
     const isEveryChildOptional = children.every(child => child.get('_isOptional'));
 
     if (isOptional && isEveryChildOptional) {
-      // As parent is optional, its completion is only used for plp
-      //   wait for the children to complete before completing
+      // As model is optional, its completion is only used for plp
+      // wait for the children to complete before completing
       completed = children.every(child => child.get(completionAttribute));
     } else if (requireCompletionOf === -1) { // a value of -1 indicates that ALL mandatory children must be completed
       completed = children.every(child => {


js/models/adaptModel.js
```diff
@@ -321,9 +321,12 @@ export default class AdaptModel extends LockingModel {
     let completed = false;
     const children = this.getAvailableChildModels();
     const requireCompletionOf = this.get('_requireCompletionOf');
+    const isParentOptional = this.get('_isOptional');
```
danielghost, line 324:
```suggestion
    const isOptional = this.get('_isOptional');
```

js/models/adaptModel.js
```diff
@@ -321,9 +321,12 @@ export default class AdaptModel extends LockingModel {
...
     const isEveryChildOptional = children.every(child => child.get('_isOptional'));
 
-    if (isEveryChildOptional) {
+    if (isParentOptional && isEveryChildOptional) {
```
danielghost, line 327:
```suggestion
    if (isOptional && isEveryChildOptional) {
```

js/models/adaptModel.js
```diff
@@ -321,9 +321,12 @@ export default class AdaptModel extends LockingModel {
...
+      // As parent is optional, its completion is only used for plp
+      //   wait for the children to complete before completing
```
danielghost, line 329:
```suggestion
      // As model is optional, its completion is only used for plp
      // wait for the children to complete before completing
```

danielghost (approved):

cahirodoherty-learningpool (approved):
👀

PR #305, opened 2023-02-21
Update: score values responsibility

### LINKED ISSUE #262: score value responsibility
Currently [tracking.js](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/tracking.js#L49) is responsible for determining what score values to submit, but I think this should probably be the responsibility of the relevant tracking plugins instead.

The `isPercentageBased` property of an assessment shouldn't really be used to determine the score format submitted, as I would argue this is more for internal calculations within assessments and their results. Depending on the configuration, using this value can lead to SCORM errors as the property now becomes tied to the SCORM version. For example, SCORM 1.2 courses where the assessment raw score value exceeds 100, will always need `isPercentageBased: true`, and will error if not. SCORM 2004 courses which can separate raw and scaled scores, would need `isPercentageBased: false`.

If `OfflineStorage` always just sends the actual raw value regardless, leaving the tracking plugins to convert these values into the relevant format, any potential configuration errors are removed and `isPercentageBased` becomes decoupled and only serves the assessment logic.

https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L227 would need to convert the raw score into a percentage. Likewise, other tracking plugins may need to change certain logic.

danielghost:
Fixes #262.

### Update
* core now simply sends the raw, min and max score values for other tracking plugins to format as required

### DIFF
--- a/js/tracking.js
+++ b/js/tracking.js
@@ -45,12 +45,6 @@ class Tracking extends Backbone.Controller {
 
   submitScore() {
     if (!this._config._shouldSubmitScore) return;
-
-    if (this._assessmentState.isPercentageBased) {
-      offlineStorage.set('score', this._assessmentState.scoreAsPercent, 0, 100);
-      return;
-    }
-
     offlineStorage.set('score', this._assessmentState.score, this._assessmentState.minScore, this._assessmentState.maxScore);
   }
 
@@ -71,7 +65,7 @@ class Tracking extends Backbone.Controller {
     if (completionData.status === COMPLETION_STATE.INCOMPLETE) {
       return;
     }
-    
+
     const canRetry = (completionData.assessment?.canRetry === true);
     if (completionData.status === COMPLETION_STATE.FAILED && canRetry) {
       return;


js/tracking.js
```diff
@@ -45,12 +45,6 @@ class Tracking extends Backbone.Controller {
 
   submitScore() {
     if (!this._config._shouldSubmitScore) return;
-
-    if (this._assessmentState.isPercentageBased) {
-      offlineStorage.set('score', this._assessmentState.scoreAsPercent, 0, 100);
-      return;
-    }
-
     offlineStorage.set('score', this._assessmentState.score, this._assessmentState.minScore, this._assessmentState.maxScore);
```
oliverfoster, line 48:
```suggestion
    offlineStorage.set('score', this._assessmentState.score, this._assessmentState.minScore, this._assessmentState.maxScore, this._assessmentState.isPercentageBased);
```

oliverfoster (approved):

eleanor-heath (approved):

cahirodoherty-learningpool:
@danielghost Merge conflict needs resolved

PR #313, opened 2023-03-03
Fix: Set drawer position on multi-extension button menu view

### LINKED ISSUE #312: Drawer opens on left side of screen on first interaction
Possibly caused by https://github.com/adaptlearning/adapt-contrib-core/issues/272

When the drawer icon is first clicked on the course the drawer is aligned to the left of the page instead of the right as expected. Interacting with the glossary when aligned to the left then forces the drawer to the right side of the screen as expected.
Also if you close the drawer and access it again in the same session then it will always appear aligned to the right.

### Your environment
* Framework v5.26.1
* Vanilla v9.4.3
* Box Menu v6.3.5

### Steps to reproduce
Create a new course
Add a page with a text component on it
Add any of the extensions that appear in the drawer - resources, glossary, etc
Ensure there are items added to the extension
Preview the course
Click on the drawer icon from either the menu or within the first page

### Expected behaviour
Drawer will open on the right side of the screen

### Actual behaviour
Drawer opens on the left of the screen when first accessed, but will then appear on the right for any subsequent access

oliverfoster:
fixes #312 

### Fix
* Drawer wasn't setting its position on multi-extension button menu view

### DIFF
--- a/js/views/drawerView.js
+++ b/js/views/drawerView.js
@@ -109,6 +109,7 @@ class DrawerView extends Backbone.View {
   }
 
   showDrawer(emptyDrawer) {
+    this.setDrawerPosition();
     this.$el.removeClass('u-display-none').removeAttr('aria-hidden');
     // Only trigger popup:opened if drawer is visible, pass popup manager drawer element
     if (!this._isVisible) {


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):

danielghost:
See https://github.com/adaptlearning/adapt-contrib-core/issues/312#issuecomment-1462066858

Issue #314, opened 2023-03-03
Add header template support for extensions


guywillis:
### Subject of the enhancement
The current implementation of the header template only supports components.

### Your environment
* Master FW / AT

### Expected behaviour
The header template should allow support for either components or extensions.

[Resources linked issue](https://github.com/adaptlearning/adapt-contrib-resources/issues/99)

PR #315, opened 2023-03-03
Fix: Add header template support for extensions (fixes: #314)

guywillis:
Fixes: #314 

### Fix
* Add support in the header template for extensions


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #322, opened 2023-03-09
Fix: Fail silently on changed() when not jsx

### LINKED ISSUE #321: Non-React components calling 'changed'
Submit of non-react components, within questionView.js - onSubmitClicked() calls 'this.changed()'. This is breaking behaviour of non-react components.

I know the goal is to have all components refactored to jsx but this is currently not 100% coverage.

### Your environment
* 5.26.1

### Steps to reproduce
Add in a non-react question component and try to submit.

### Expected behaviour
Component should still submit

### Actual behaviour
Breaking and spitting out Uncaught Error: Cannot call changed on a non-react view

oliverfoster:
fixes #321 

### Fix
* Allow changed to be called and to fail silently on non-jsx views.

### DIFF
--- a/js/views/adaptView.js
+++ b/js/views/adaptView.js
@@ -86,9 +86,7 @@ class AdaptView extends Backbone.View {
    */
   changed(eventName = null) {
     if (this._jsxIgnoreChanges !== 0) return;
-    if (!this.isJSX) {
-      throw new Error('Cannot call changed on a non-react view');
-    }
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

eleanor-heath (approved):

PR #329, opened 2023-03-13
New: Add icon span to instruction element in header.jsx (fixes #328)

### LINKED ISSUE #328: Add intruction error icons
The Narrative component needs to highlight the instruction text with error styles and an icon. This could be applied to other components as well or even articles and blocks.

This implementation should hide the icon from screen readers and, because of this, is preferred to using a pseudo-selector for the icon.

https://github.com/adaptlearning/adapt-contrib-narrative/pull/250

Shout out to @kirsty-hames for this idea.

swashbuck:
Fixes #328 

### New
* Adds a `span` element with an `icon` class. This can be used to show an icon when, for instance, the instruction text should be highlighted as an error.

### Testing
Can be tested along with https://github.com/adaptlearning/adapt-contrib-narrative/pull/250 .
1. Add a Narrative component
2. On mobile, click the Next button without first clicking on the strapline button.

### Notes

- All styling for this error message / icon currently lives in the Narrative component's Less. We could consider moving it to the theme or core styles, though.

### DIFF
--- a/templates/header.jsx
+++ b/templates/header.jsx
@@ -73,6 +73,7 @@ export default function Header(props) {
 
         {sizedInstruction &&
         <div className={prefixClasses(classNamePrefixes, ['__instruction'])}>
+          <span className="icon" aria-hidden="true" />
           <div className={prefixClasses(classNamePrefixes, ['__instruction-inner'])} dangerouslySetInnerHTML={{ __html: compile(sizedInstruction, props) }}>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

kirsty-hames (changes requested):
Thanks for picking this up @swashbuck. 

Whilst we have some components that haven't been converted to JSX, we'll also need to update _component.hbs_ too.

oliverfoster (approved):

kirsty-hames (approved):

PR #336, opened 2023-03-27
New: added successful reset event trigger (fixes #335)

### LINKED ISSUE #335: AdaptModel reset trigger
Currently it isn't easy to know when an incomplete model with children is reset. We can listen to a `change:_isInteractionComplete` or  `change:_isComplete` to know when the status has been changed, but this won't inform us of reset being called when the parent status is unchanged.

For plugins which modify child models of an associated parent, or work with a collection of models like an assessment, it would be useful to know when a reset has been called on that parent model regardless of whether the completion status changed.

### Proposed solution
Add `this.trigger('reset', this);` following a successful reset.

danielghost:
Fixes #335

### New
* Added a successful `'reset'` event trigger for AdaptModel

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -228,20 +228,19 @@ export default class AdaptModel extends LockingModel {
    */
   reset(type = 'hard', canReset = this.get('_canReset')) {
     if (!canReset) return false;
+    let resetData = {
+      _isEnabled: true,
+      _isInteractionComplete: false
+    };
     switch (type) {
       case 'hard':
       case true:
-        this.set({
-          _isEnabled: true,
-          _isComplete: false,
-          _isInteractionComplete: false
+        Object.assign(resetData, {
+          _isComplete: false
         });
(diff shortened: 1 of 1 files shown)


js/models/adaptModel.js
```diff
@@ -228,23 +228,17 @@ export default class AdaptModel extends LockingModel {
...
-        return true;
-    }
-    return false;
+    const resetData = {
+      _isEnabled: true,
+      _isInteractionComplete: false
+    };
+    const isHardReset = (type === 'hard' || type === true);
+    const isSoftReset = (type === 'soft');
+    if (!isHardReset && !isSoftReset) return false
```
cahirodoherty-learningpool, line 237: Minor: You can move the resetData object to after this early return. It isn't used otherwise

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
Tested with latest FW and core, working as expected.

PR #339, opened 2023-03-30
New: Added Navigation Button API

### LINKED ISSUE #338: Navigation Button API
### Subject of the issue
Allow extensions to define buttons and add them to the navigation bar rather than allowing DOM injection.

* Backward compatible with injection
* Text label based upon the button aria label
* Ordering
* Icons
* Show/hide text

oliverfoster:
fixes #338 

### New
* `NavigationButtonModel` To hold the properties for each button
* `NavigationButtonView` For extensions to make their own buttons and for legacy injected button management
* `NavigationModel` To hold the navigation configuration
* All model updates will be reflected in the DOM.

General API layout:
```js
import navigation from 'core/js/navigation';
navigation.model.set('_showLabel', true); // Show all labels
const instance = navigation.getButton('back');
instance.set('_order', 400); // Move the back button to the right
instance.set('text', 'New Label');; // Change the back button label text
navigation.removeButton(instance);
// Remove all buttons
navigation.buttons.forEach(button => navigation.removeButton(button));
```

To make a home button:

image

```js
import Adapt from 'core/js/adapt';
import navigation from 'core/js/navigation';
import NavigationButtonModel from 'core/js/models/NavigationButtonModel';
import NavigationButtonView from 'core/js/views/NavigationButtonView';

// Extend the model and view classes to add own behaviour

Adapt.on('navigation:ready', () => {
  const model = new NavigationButtonModel({
    _id: 'home',
    _order: 0,
    _event: 'homeButton',
    _showLabel: null,
    _classes: '',
    _iconClasses: 'icon-medal',
    _role: 'link',
    ariaLabel: 'Home',
    text: '{{ariaLabel}}'
  });
  const view = new NavigationButtonView({ model });
  navigation.addButton(view);
});
```

`_eventName` has some default behaviour at `"backButton"`, `"homeButton"`, `"parentButton"`, `"skipNavigation"` and `"returnToStart"`.

### Testing
In the console it is possible to use:
1. `require('core/js/navigation').model.set('_showLabel', true);`
2. `require('core/js/navigation').getButton('back').set('_order', 400);`
3. `require('core/js/navigation').getButton('back').set('text', 'New Label');`

### DIFF
--- a/js/models/NavigationModel.js
+++ b/js/models/NavigationModel.js
@@ -1,0 +1,13 @@
+import LockingModel from 'core/js/models/lockingModel';
+
+export default class NavigationModel extends LockingModel {
+
+  defaults() {
+    return {
+      _navigationAlignment: 'top',
+      _isBottomOnTouchDevices: false,
+      _showLabel: false
+    };
+  }
+
+}
--- a/js/views/NavigationButtonView.js
+++ b/js/views/NavigationButtonView.js
@@ -1,0 +1,151 @@
+import Adapt from 'core/js/adapt';
+import wait from 'core/js/wait';
+import { compile, templates } from 'core/js/reactHelpers';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import router from 'core/js/router';
+import startController from 'core/js/startController';
+import a11y from 'core/js/a11y';
+import location from 'core/js/location';
+
+export default class NavigationButtonView extends Backbone.View {
+
+  events() {
+    return {
+      click: 'triggerEvent'
+    };
+  }
+
+  className() {
+    return '';
+  }
+
+  attributes() {
+    const attributes = this.model.toJSON();
+    return {
+      'data-order': attributes._order,
+      'data-event': attributes._event
+    };
+  }
+
+  initialize({ el }) {
+    if (el) {
+      this.isInjectedButton = true;
+    } else {
+      this.isJSX = (this.constructor.template || '').includes('.jsx');
+    }
+    this._classSet = new Set(_.result(this, 'className').trim().split(/\s+/));
+    this._attributes = _.result(this, 'attributes');
+    this.listenTo(this.model, 'change', this.changed);
+    this.render();
+  }
+
+  static template() {
+    return 'navigation-button.jsx';
+  }
+
+  render() {
+    if (this.isInjectedButton) {
+      this.changed();
+    } else if (this.isJSX) {
+      this.changed();
+    } else {
+      const data = this.model.toJSON();
+      data.view = this;
+      const template = Handlebars.templates[this.constructor.template];
+      this.$el.html(template(data));
+    }
+    return this;
+  }
+
+  updateViewProperties() {
+    const classesToAdd = _.result(this, 'className').trim().split(/\s+/);
+    classesToAdd.forEach(i => this._classSet.add(i));
+    const classesToRemove = [ ...this._classSet ].filter(i => !classesToAdd.includes(i));
+    classesToRemove.forEach(i => this._classSet.delete(i));
+    Object.keys(this._attributes).forEach(name => this.$el.removeAttr(name));
+    Object.entries(_.result(this, 'attributes')).forEach(([name, value]) => this.$el.attr(name, value));
+    this.$el.removeClass(classesToRemove).addClass(classesToAdd);
+  }
+
+  injectLabel() {
+    const textLabel = this.$el.find('> .label');
+    const ariaLabel = this.$el.attr('aria-label') ?? this.$el.find('.aria-label').text();
+    const text = this.model.get('text');
+    const output = compile(text ?? '', { ariaLabel });
+    if (!textLabel.length) {
+      this.$el.append(`<span class="label" aria-hidden="true">${output}</span>`);
+      return;
+    }
+    textLabel.html(output);
+  }
+
+  /**
+   * Re-render a react template
+   * @param {string} eventName=null Backbone change event name
+   */
+  changed(eventName = null) {
+    if (typeof eventName === 'string' && eventName.startsWith('bubble')) {
+      // Ignore bubbling events as they are outside of this view's scope
+      return;
+    }
+    if (this.isInjectedButton) {
+      this.updateViewProperties();
+      this.injectLabel();
+      return;
+    }
+    if (!this.isJSX) {
+      this.updateViewProperties();
+      return;
+    }
+    const props = {
+      // Add view own properties, bound functions etc
+      ...this,
+      // Add model json data
+      ...this.model.toJSON(),
+      // Add globals
+      _globals: Adapt.course.get('_globals')
+    };
+    const Template = templates[this.constructor.template.replace('.jsx', '')];
+    this.updateViewProperties();
+    ReactDOM.render(<Template {...props} />, this.el);
+  }
+
+  triggerEvent(event) {
+    event.preventDefault();
+    const currentEvent = $(event.currentTarget).attr('data-event');
+    if (!currentEvent) return;
+    Adapt.trigger('navigation:' + currentEvent);
+    switch (currentEvent) {
+      case 'backButton':
+        router.navigateToPreviousRoute();
+        break;
+      case 'homeButton':
+        router.navigateToHomeRoute();
+        break;
+      case 'parentButton':
+        router.navigateToParent();
+        break;
+      case 'skipNavigation':
+        a11y.focusFirst('.' + location._contentType);
+        break;
+      case 'returnToStart':
+        startController.returnToStartLocation();
+        break;
+    }
+  }
+
+  remove() {
+    this._isRemoved = true;
+    this.stopListening();
+    wait.for(end => {
+      if (this.isJSX) {
+        ReactDOM.unmountComponentAtNode(this.el);
+      }
+      super.remove();
+      end();
+    });
+    return this;
+  }
+
+}
--- a/js/views/navigationView.js
+++ b/js/views/navigationView.js
@@ -1,20 +1,16 @@
 import Adapt from 'core/js/adapt';
-import a11y from 'core/js/a11y';
-import location from 'core/js/location';
-import router from 'core/js/router';
-import startController from 'core/js/startController';
+import device from 'core/js/device';
 import _ from 'underscore';
+import NavigationButtonView from './NavigationButtonView';
+import NavigationButtonModel from '../models/NavigationButtonModel';
 
 class NavigationView extends Backbone.View {
 
   className() {
-    return 'nav';
-  }
-
-  events() {
-    return {
-      'click [data-event]': 'triggerEvent'
-    };
+    return [
+      'nav',
+      this.model?.get('_showLabel') && 'show-label'
+    ].join(' ');
   }
 
   attributes() {
@@ -23,11 +19,48 @@ class NavigationView extends Backbone.View {
     };
   }
 
+  get buttons() {
+    return (this._buttons = this._buttons || []);
+  }
+
+  set buttons(value) {
+    this._buttons = value;
+  }
+
   initialize() {
     _.bindAll(this, 'sortNavigationButtons');
+    this._classSet = new Set(_.result(this, 'className').trim().split(/\s+/));
+  }
+
+  start(model) {
+    this.model = model;
+    this.listenTo(model, 'change', this.update);
     this.listenToOnce(Adapt, 'courseModel:dataLoading', this.remove);
-    this.listenTo(Adapt, 'router:menu router:page', this.hideNavigationButton);
+    this.listenTo(Adapt, {
+      'router:contentObject': this.hideNavigationButton,
+      'adapt:preInitialize device:resize': this.onDeviceResize
+    });
+    this.update();
     this.preRender();
+  }
+
+  update() {
+    this.updateViewProperties();
+    this.onDeviceResize();
+  }
+
+  onDeviceResize() {
+    let {
+      _navigationAlignment,
+      _isBottomOnTouchDevices
+    } = this.model.toJSON();
+    const $html = $('html');
+    $html.addClass('is-nav-top');
+    const isBottomOnTouchDevices = (device.touch && _isBottomOnTouchDevices);
+    if (isBottomOnTouchDevices) _navigationAlignment = 'bottom';
+    $html
+      .removeClass('is-nav-top is-nav-bottom')
+      .addClass('is-nav-' + _navigationAlignment);
   }
 
   preRender() {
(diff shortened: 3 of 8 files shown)


swashbuck:
Created https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/174 for Page Level Progress style changes.

swashbuck:
@oliverfoster I've added a class called `has-label-bottom` that will move the labels below the icons. This should be an option at both the course level and the extension level.

Are two text label position options enough? Default would be label on the right, and the new option would be label under the icon. Would it be better to have something like `_labelPosition` that can be set to `top`, `left`, `right` or `bottom`?

oliverfoster:
> @oliverfoster I've added a class called `has-label-bottom` that will move the labels below the icons. This should be an option at both the course level and the extension level.
> 
> 
> 
> 
> 
> **Question:** Are two text label position options enough? Default would be label on the right, and the new option would be label under the icon. Would it be better to have something like `_labelPosition` that can be set to `top`, `left`, `right` or `bottom`?

I'm sure there will be an rtl requirement at minimum. @guywillis positioning controls at a course level?

swashbuck:
> I'm sure there will be an rtl requirement at minimum. @guywillis positioning controls at a course level?

@oliverfoster Ok, I can create classes for `has-label-bottom`, `has-label-top`, `has-label-left`, and `has-label-right`. That should cover everything. Then, we'll need a new setting/property for the model.

swashbuck:
> Hide text below medium

I've implemented this in the Less. If we need to give devs more control (which might be overkill), we could add another `course.json` property.

Property name: `_showLabelAtWidth`
- _empty_ - Always show the label
- `small` - Show at small width and up
- `medium` - Show at medium width and up
- `large` - Show at large width and up

This would add a class to `.nav` like `show-label-small` which can then be used to show the labels at certain breakpoints.

oliverfoster:
> > Hide text below medium
> 
> I've implemented this in the Less for now. If we need to give devs more control (which might be overkill), we could add another `course.json` property.
> 
> Property name: `_showLabelAtWidth`
> 
> * _empty_ - Always show the label
> * `small` - Show at small width and up
> * `medium` - Show at medium width and up
> * `large` - Show at large width and up
> 
> This would add a class to `.nav` like `show-label-small` which can then be used to show the labels at certain breakpoints.
> 
> ```
>   "_navigation": {
>     "_showLabel": true,
>     "_showLabelAtWidth": "medium"
>   }
> ```
I'd like to get @guywillis input on this bit as it was his biggest concern, along with a review of the styling.

oliverfoster:
I think, in lieu of an option in the JSON, these classes should be sufficient for custom styling. 

What should be the default in either case?

swashbuck:
Positioning and breakpoint settings have been added at the top level.

_course.json_
```
  "_navigation": {
    "_showLabel": true,
    "_showLabelAtWidth": "small|medium|large",
    "_labelPosition": "left|right|top|bottom"
  },
```

swashbuck:
Schemas have been added for the new options.

eleanor-heath (approved):

guywillis:
I'll preface with a comment to say I genuinely think navigation labels are a great idea but I think we need to take a bit more time to finesse and future proof the concept. 

## Issues

### Labels are linked to aria labels
Linking labels to the aria label text entry might sound like a nice shortcut but I can envisage a situation where one may want to only attach labels to a particular button and therefore remove key screen reader aria labels from other buttons to achieve this outcome. 

Ideally, there would be properties for navigation labels that are separate from screen reader aria labels. Potentially even plugin level config to enable / disable labels.

### Labels do not display using `"_showLabelAtWidth": "large",`
I think this could potentially be a fairly simple bug but nevertheless it doesn't display currently when using this setting.

### Rendering an empty `<span>` within the `.nav__btn` element causes icon only buttons to have a visual error
Extra space is rendered within the navigational buttons due to an empty span tag. This causes an issue when either removing the aria label and thus label from a button in core or using a 3rd party plugin that doesn't yet support the navigation label feature. This causes the button to no longer be square. 

Screenshot 2023-04-18 at 14 19 55

### Length of navigation labels
As they stand the length of the labels are fairly long and leave very little room for other elements often found in the navigational area such as logos, titles, or additional plugins. The screenshot below shows limited space available on a 1200px wide screen - it is not unusual for devices to be smaller than this. 

Screenshot 2023-04-18 at 16 04 12

### Breakpoints for `_showLabelAtWidth`
The breakpoints for the show / hide label functionality seem rather arbitrary and not necessarily the best for the content shown within - triggering too early / too late etc. I'm acutely aware the current implementation may be due to limitations but ideally a more content aware solution could be used.

### Incorrect calculation of wrapper padding with `_labelPosition: "bottom"`
It does not look like the `@navigation-height;` variable is taking into account the increase in height caused by the labels. 

Screenshot 2023-04-18 at 14 23 16

### Label positioning using `order`
Arguably, we should not be using this CSS as it falls foul for WCAG 2.0 guidelines relating to DOM order matching the visual order - https://www.w3.org/TR/WCAG20-TECHS/C27.html 

If we want to support these alternative label layouts we should be editing the DOM. 

### White text labels fail AA contrast on navigation buttons
As described, the white text labels fail AA colour contrast when displayed on the navigational buttons. We're likely need to adjust the Vanilla theme colours to resolve the issue. 

Screenshot 2023-04-18 at 15 46 20

### Missing RTL support
There are no margin inversions to support RTL. Please can these be added in.

## Code adjustments

### Using `>` selector
I'm not sure I entirely see the necessity of using the `>` CSS selector - could you shed some light on this please?

### Reduce unnecessary nesting
There's lots of nesting in the new CSS which could be optimised by adding to a mixin or moving selectors onto the same line. Preferably, the CSS is defined in less than 3 indentations.

## Other

### Editing in browser causes DOM refresh making it increasingly problematic to debug
I figure this is due to the code change and not something that could be amended but regardless, when I try to attach some styling to an element within the navigation area with some inline styling the DOM refreshes. This makes the process rather unfriendly to developers.

In the example video I was trying to attach `background-color: red;` to the drawer navigation button.
https://user-images.githubusercontent.com/11569678/232810254-012f4a4a-1af2-45a1-bd50-9ac2f97b24a6.mov

## Testing

I have been using the following settings and editing the values of `_showLabelAtWidth` and `_labelPosition` whilst testing:

```
  "_navigation": {
    "_isDefaultNavigationDisabled": false,
    "_navigationAlignment": "top",
    "_isBottomOnTouchDevices": false,
    "_showLabel": true,
    "_showLabelAtWidth": "medium",
    "_labelPosition": "right"
  },
```

oliverfoster:
> ### Labels are linked to aria labels
> Linking labels to the aria label text entry might sound like a nice shortcut but I can envisage a situation where one may want to only attach labels to a particular button and therefore remove key screen reader aria labels from other buttons to achieve this outcome.
> 
> Ideally, there would be properties for navigation labels that are separate from screen reader aria labels. Potentially even plugin level config to enable / disable labels.

Yes. This is as planned. The button models have both `ariaLabel` and `text` attributes, `text` currently defaults to `{{ariaLabel}}` for buttons which don't supply their own config. Those will come once the api is in, plp etc will need updating and drawer needs fixing with multiple sub extensions, as resources forcefully changes the button aria-label currently.

> ### Length of navigation labels
> As they stand the length of the labels are fairly long and leave very little room for other elements often found in the navigational area such as logos, titles, or additional plugins. The screenshot below shows limited space available on a 1200px wide screen - it is not unusual for devices to be smaller than this.

Similar to above. The labels as they are are not the final labels, only a rendering of the current aria-labels.

> ### Label positioning using `order`
> Arguably, we should not be using this CSS as it falls foul for WCAG 2.0 guidelines relating to DOM order matching the visual order - https://www.w3.org/TR/WCAG20-TECHS/C27.html
> 
> If we want to support these alternative label layouts we should be editing the DOM.

The button's aria label will read, not the text supplied for the visual label. We don't need to worry about dom order as a result, the `aria-hidden=true` on the labels has it sorted.

> ## Other
> ### Editing in browser causes DOM refresh making it increasingly problematic to debug
> I figure this is due to the code change and not something that could be amended but regardless, when I try to attach some styling to an element within the navigation area with some inline styling the DOM refreshes. This makes the process rather unfriendly to developers.
> 
> In the example video I was trying to attach `background-color: red;` to the drawer navigation button. https://user-images.githubusercontent.com/11569678/232810254-012f4a4a-1af2-45a1-bd50-9ac2f97b24a6.mov

Yup, duly noted, this is expected due to necessary backward compatibility. The nav bar listens for DOM changes in order to keep the model inline with the dom, for injected buttons, where they change. To get around this for the time being, you'll have to add a style to an existing class, instead of adding a `style` attribute. In future each button will have configurable custom classes, which you can simulate using: 
```js
require('core/js/navigation').getButton('back').set('_classes', 'new-class');
```

swashbuck:
@guywillis Thanks for the thorough review!

**Labels do not display using "_showLabelAtWidth": "large",**
This should now be fixed.

**Rendering an empty <span> within the .nav__btn element causes icon only buttons to have a visual error**
Will need to investigate this.

**Breakpoints for _showLabelAtWidth**
This aligns with current Adapt breakpoints (small, medium, large). It should be up to the developer/designer to ensure that the label text isn't too long. Otherwise, what would you recommend in lieu of standard Adapt breakpoints?

**Incorrect calculation of wrapper padding with _labelPosition: "bottom"**
Will need to investigate this.

**White text labels fail AA contrast on navigation buttons**
Should be covered by https://github.com/adaptlearning/adapt-contrib-vanilla/pull/415

**Missing RTL support**
Can this not be addressed by the _labelPosition setting?

**Using > selector / Reduce unnecessary nesting**
The Less could probably use some optimizing. Agree that > selector is probably unnecessary and too specific.

swashbuck:
@guywillis I believe I've addressed all your points above. Let me know what you think.

guywillis:
Thank you for the swift responses @oliverfoster and @swashbuck

**Labels do not display using "_showLabelAtWidth": "large",**
Can confirm fixed.

**Rendering an empty within the .nav__btn element causes icon only buttons to have a visual error**
Understood. To confirm that it should be classified as an all on or off option at present until such a time where plugins take control of their label display.

**Breakpoints for _showLabelAtWidth**
Fair points made. As an aside we really need to improve Adapt's breakpoints. 

**Incorrect calculation of wrapper padding with _labelPosition: "bottom"**
I'd be hesitant to go down the inline js route but if that's the solution so be it. I'd have more concern over assuming a single line text especially for responsive and translation views.

Would CSS variables be a potential solution to this issue? In a similar vein to the `--adapt-viewport-height`:
[Link One](https://github.com/adaptlearning/adapt-contrib-core/blob/master/less/_defaults/_variables.less#L2)
[Link Two](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/device.js#L105)

**Missing RTL support**
Comes down to automation vs manual I guess. If I am translating a course from English to Arabic would I, as the developer, be expected to go in and change the layout before / after translation? Potentially confusing as other elements of Adapt are automatically converted. 

**Using > selector / Reduce unnecessary nesting**
Thank you for the improvements. I shall post a couple of suggestions and be keen to hear your thoughts on them

less/core/nav.less
```diff
@@ -24,16 +24,60 @@ html.is-nav-bottom #wrapper {
...
+    .show-nav-label(medium);
+    .show-nav-label(large);
+  }
+
+  &.has-label-right &__btn-label {
+    margin: 0 0 0 0.5rem;
+  }
+
+  &.has-label-bottom {
+    .nav__btn {
```
guywillis, line 56: Keen to get your thoughts on two entries rather than nesting:

```
&.has-label-bottom &__btn {}

&.has-label-bottom &__btn-label {}
```

less/core/nav.less
```diff
@@ -24,16 +24,60 @@ html.is-nav-bottom #wrapper {
...
+  }
+
+  &.show-label {
+    .show-nav-label(small);
+    .show-nav-label(medium);
+    .show-nav-label(large);
+  }
+
+  &.has-label-right &__btn-label {
+    margin: 0 0 0 0.5rem;
```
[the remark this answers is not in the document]
kirsty-hames: Would it make sense to use `(@icon-padding / 2)` here? It's the same value (.5rem) but it will keep the margin relative to the button padding. It's also consistent with the [pageNav](https://github.com/cgkineo/adapt-pageNav/blob/b7cc4a7864b1b3e8ce0a54d1ffe87e3cd3a4efd1/less/pageNav.less#L51) plugin (not a core plugin I know) but maintains consistency across navigation UI.

kirsty-hames:
> **Missing RTL support**
> Comes down to automation vs manual I guess. If I am translating a course from English to Arabic would I, as the developer, be expected to go in and change the layout before / after translation? Potentially confusing as other elements of Adapt are automatically converted.

I agree, I would expect this to be automated. I would also expect the default layout to inherit from the language direction e.g [ icon / text] for LTR and [ text / icon ] for RTL.

I'd imagine the only reason why you wouldn't want the icon/text display to follow the language direction is if it's specific to a brand etc. In which case you would want to maintain this regardless of language. Is it worth including a default `_labelPosition: auto`?

`_labelPosition: auto` - default display, inherits from lang direction [ icon / text] for LTR and [ text / icon ] for RTL
`_labelPosition: left` - label displays left of icon regardless of lang direction
`_labelPosition: right` - label displays right of icon regardless of lang direction

For consistency of config, we already use the auto/left/right approach for the [drawer position](https://github.com/adaptlearning/adapt-contrib-core/pull/219).

less/core/nav.less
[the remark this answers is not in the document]
swashbuck: Thanks, both. I like the @icon-padding idea here.

> The 0 from the 0.5 value can also be dropped.
In general, I would be hesitant to change this, but maybe it's personal preference. If this is the convention that's most widely used in Adapt Less, then it makes sense to align with that standard. In VSC, I enable an error for "less.lint.zeroUnits / No unit for zero needed" since I think it makes it more readable. You're far less likely to confuse "0.5" with "5" as opposed to not using a leading zero like ".5"

less/core/nav.less
```diff
@@ -24,16 +24,60 @@ html.is-nav-bottom #wrapper {
...
+
+    .nav__btn-label {
+      order: -1;
+      margin: 0 0 0.5rem 0;
+    }
+  }
+
+  &.has-label-left &__btn-label {
+    order: -1;
+    margin: 0 0.5rem 0 0;
```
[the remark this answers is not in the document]
swashbuck: Done. Updated margin syntax and switched to @icon-padding / 2

less/core/nav.less
```diff
@@ -24,16 +24,60 @@ html.is-nav-bottom #wrapper {
...
+  }
+
+  &.has-label-top {
+    .nav__btn {
+      flex-direction: column;
+    }
+
+    .nav__btn-label {
+      order: -1;
+      margin: 0 0 0.5rem 0;
```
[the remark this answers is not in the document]
swashbuck: Done. Updated margin syntax and switched to @icon-padding / 2

less/core/nav.less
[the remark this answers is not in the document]
swashbuck: Done. Updated margin syntax and switched to @icon-padding / 2

less/core/nav.less
swashbuck: I'm fine with the nesting, but I don't have strong feelings either way.

Nesting allows for code folding in an editor, which is nice, and it also groups the styles for each label position in the same code block.

swashbuck:
> **Incorrect calculation of wrapper padding with _labelPosition: "bottom"** I'd be hesitant to go down the inline js route but if that's the solution so be it. I'd have more concern over assuming a single line text especially for responsive and translation views.
> 
> Would CSS variables be a potential solution to this issue? In a similar vein to the `--adapt-viewport-height`: [Link One](https://github.com/adaptlearning/adapt-contrib-core/blob/master/less/_defaults/_variables.less#L2) [Link Two](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/device.js#L105)

@guywillis I love the idea of CSS variables here. I can work on this.

> I agree, I would expect this to be automated. I would also expect the default layout to inherit from the language direction e.g [ icon / text] for LTR and [ text / icon ] for RTL.

@kirsty-hames Yes, this makes sense. I will add an "auto" option and make that the default position value.

swashbuck:
@guywillis I've added a new `--adapt-navigation-height` CSS variable that will be used for wrapper and drawer padding. This value will be updated via device.js.

I've also reworked the drawer toolbar a bit so that the toolbar is the same height as the nav and the buttons are centered vertically.

less/_defaults/_variables.less
```diff
@@ -79,10 +80,6 @@
 // --------------------------------------------------
 @component-padding-hoz: 0.5rem;
 
-// Navigation height
-// --------------------------------------------------
-@navigation-height: @icon-size + (@icon-padding * 2);
```
oliverfoster, line 84: Probably wise to keep this default in for the time being? With a deprecated comment.

less/core/nav.less
guywillis: I can certainly see the benefit of this approach.

less/core/nav.less
```diff
@@ -24,16 +24,65 @@ html.is-nav-bottom #wrapper {
...
+    display: none;
+  }
+
+  &.show-label {
+    .show-nav-label(small);
+    .show-nav-label(medium);
+    .show-nav-label(large);
+  }
+
+  .dir-ltr &.has-label-auto &__btn-label,
```
guywillis, line 51: Although this selector shares the same styling as `&.has-label-right &__btn-label` should it, and the rtl varient, be defined separately?

guywillis:
> @guywillis I've added a new `--adapt-navigation-height` CSS variable that will be used for wrapper and drawer padding. This value will be updated via device.js.
> 
> I've also reworked the drawer toolbar a bit so that the toolbar is the same height as the nav and the buttons are centered vertically.

It seems the CSS variable has an undefined value when first loading the course:

Screenshot 2023-04-24 at 14 45 36

If I resize the screen the value is updated with the correct value:

Screenshot 2023-04-24 at 14 46 18

Otherwise the CSS variable implementation looks to be working great.

The `@navigation-height` variable probably needs to remain in both Core and Vanilla until a breaking change in the future where we can remove it.

The drawer toolbar amends also look great.

swashbuck:
> It seems the CSS variable has an undefined value when first loading the course:

@guywillis Good catch. Will you test this again? I've moved the initial `setNavigationHeight()` call to `navigationView:postRender`. This should ensure that the navigation element exists and has a height before polling it for the height value.

guywillis:
> > It seems the CSS variable has an undefined value when first loading the course:
> 
> @guywillis Good catch. Will you test this again? I've moved the initial `setNavigationHeight()` call to `navigationView:postRender`. This should ensure that the navigation element exists and has a height before polling it for the height value.

The change looks to work great, thank you.

One final comment to raise from myself. The setting `_showLabelAtWidth` assumes the course creator will always want to hide the labels at a certain adapt breakpoint - what if I want to always show the labels?

less/core/nav.less
guywillis: We do, yes. However, we usually nest the rtl style under the selector it is styling for clarity.

[button example](https://github.com/adaptlearning/adapt-contrib-core/blob/master/less/core/button.less#L21)

The rtl styles above clearly pertain to the button action selector.

The highlighted code above is good from a code optimisation perspective as it's not repeating styles but it does make the code harder to read as it's combining two unrelated selectors together `.dir-ltr &.has-label-auto &__btn-label` and `&.has-label-right &__btn-label`

guywillis:
I knew I shouldn't have said one final comment.. 

Is it expected to have the label `nav__btn-label` appear in non button related navigational elements - be it 3rd party or core?

Screenshot 2023-04-25 at 09 31 38

oliverfoster:
> I knew I shouldn't have said one final comment..
> 
> Is it expected to have the label `nav__btn-label` appear in non button related navigational elements - be it 3rd party or core?
> 
> Screenshot 2023-04-25 at 09 31 38

Yeas, currently any element that gets injected into the navigation bar, has a label injected. Except the spacer
https://github.com/adaptlearning/adapt-contrib-core/blob/d5cb6c1ff8f79313108b54f99ab66a9d3fe60212/js/views/navigationView.js#L132-L169

I could make it buttons only if preferable?

guywillis:
> Yeas, currently any element that gets injected into the navigation bar, has a label injected. Except the spacer
> I could make it buttons only if preferable?

I can certainly see benefits of both approaches. I didn't have a set solution in mind when making the comment, more of an open question.

swashbuck:
> One final comment to raise from myself. The setting _showLabelAtWidth assumes the course creator will always want to hide the labels at a certain adapt breakpoint - what if I want to always show the labels?

@guywillis I've modified the styles a bit and added a new option called "any" that will show the labels at any size. Labels can still be hidden completely at the global level with `_navigation._showLabel: false` or at the plugin level (ex. PLP https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/180 ).

swashbuck:
Just for my own sanity, these are all of the Navigation Button API-related PRs:

[​https://github.com/adaptlearning/adapt-contrib-core/pull/339](https://github.com/adaptlearning/adapt-contrib-core/pull/339)
[​https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/174](https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/174)
[​https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/180](https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/180)
[​https://github.com/adaptlearning/adapt-contrib-trickle/pull/188](https://github.com/adaptlearning/adapt-contrib-trickle/pull/188)
[​https://github.com/adaptlearning/adapt-contrib-vanilla/pull/415](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/415)

js/models/NavigationModel.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default class NavigationModel extends LockingModel {
+
+  defaults() {
+    return {
+      _navigationAlignment: 'top',
+      _isBottomOnTouchDevices: false,
+      _showLabel: false,
+      _showLabelAtWidth: 'medium',
+      _labelPosition: 'right'
```
guywillis, line 11: Should we update this default to be auto?

js/views/navigationView.js
```diff
@@ -1,20 +1,18 @@
...
-
-  events() {
-    return {
-      'click [data-event]': 'triggerEvent'
-    };
+    return [
+      'nav',
+      this.model?.get('_showLabel') === true ? 'show-label' : 'hide-label',
+      `show-label-${this.model?.get('_showLabelAtWidth') || 'medium'}`,
+      `has-label-${this.model?.get('_labelPosition') || 'right'}`
```
guywillis, line 14: Should we update this default to be auto?

guywillis (approved):
👍 

Couple of comments to adjust defaults but otherwise I'm happy - really nice work

less/core/nav.less
swashbuck: @guywillis Let me know if nesting the rtl/ltr styles is now easier to read. See [635c8c7](https://github.com/adaptlearning/adapt-contrib-core/pull/339/commits/635c8c72e329f1e5581e69b84c6738bd730ae2ff)

guywillis: It does, thank you

swashbuck (approved):
👍

oliverfoster:
One more approval and we're ready to merge.

js/views/NavigationButtonView.js
```diff
@@ -0,0 +1,174 @@
...
+
+export default class NavigationButtonView extends Backbone.View {
+
+  tagName() {
+    return 'button';
+  }
+
+  events() {
+    return {
+      click: 'triggerEvent'
```
eleanor-heath, line 19: missing semicolon here

js/views/navigationView.js
```diff
@@ -24,17 +26,54 @@ class NavigationView extends Backbone.View {
...
+    this.onDeviceResize();
+  }
+
+  onDeviceResize() {
+    let {
+      _navigationAlignment,
+      _isBottomOnTouchDevices
+    } = this.model.toJSON();
+    const $html = $('html');
+    $html.addClass('is-nav-top');
```
eleanor-heath, line 69: why is the `is-nav-top` class added here if it is removed immediately below in line 73?

templates/navButton.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { classes, compile } from 'core/js/reactHelpers';
+
+export default function NavButton(props) {
+  const {
+    text,
+    _iconClasses
+  } = props;
+  return (
+    <React.Fragment>
```
eleanor-heath, line 10: Not sure what the preference here is but we could use the newer syntax `<></>` here over `<React.Fragment></React.Fragment>`

js/views/navigationView.js
oliverfoster: Corrected

Issue #334, opened 2023-03-24
Incorrect Not Final feedback should use Incorrect Final as a fallback


cahirodoherty-learningpool:
### Subject of the issue/enhancement/features
Incorrect Not Final feedback is not using Incorrect Final as a fallback, even though the help text states this

### Your environment
* FW 5.28.4

### Steps to reproduce
Create an assessment with multiple attempts
Set feedback for Incorrect Final
Do not set feedback for Incorrect Not Final
Preview the course and fail the assessment

### Expected behaviour
Incorrect Final feedback is displayed

### Actual behaviour
Only Feedback Title is displayed

PR #342, opened 2023-04-03
Fix: Feedback incorrect not final should use fallback

cahirodoherty-learningpool:
#334 

### Fix
* Feedback incorrect not final should fall back to feedback incorrect final value if not set itself

### Testing
1. Create a course with question component and tutor
2. Allow multiple attempts on question, but do not populate incorrect not final value
3. Feedback body is empty


eleanor-heath (approved):

oliverfoster (approved):

swashbuck (approved):
👍

PR #343, opened 2023-04-04
Update: Added change vs initial language selection attribute

### LINKED ISSUE #188: the event "app:languageChanged" is triggered on initial selection of language (and on resuming a course)
Which can lead to unexpected behaviour in certain circumstances.

One thing to consider: should the event be triggered on initial selection/resume of a language that's different to `_defaultLanguage` language as configured in config.json - or should it _never_ be triggered on initial selection/resume, so only triggered when actually _changing_ the language mid-course?

Equally, we could add a new `"app:languageSelected"` event to differentiate between initial selection and a mid-course change of language.

Or, in v6, change over completely to `"app:languageSelected"` and, on initial selection/resume, do:
```js
Adapt.trigger('app:languageSelected', newLanguage);
```
and, if the learner has changed language mid-course, do:
```js
Adapt.trigger('app:languageSelected', newLanguage, oldLanguage);
```
?

oliverfoster:
fixes #188 

### Update
* The `app:languageChange` event now has two attributes when the language is changing and not just loading.
`app:languageChanged(selectedLanguage)` Will trigger when the course loads
`app:languageChanged(newLanguage, previousLanguage)` Will trigger when the language is changed

### DIFF
--- a/js/data.js
+++ b/js/data.js
@@ -79,6 +79,8 @@ class Data extends AdaptCollection {
 
   async onLanguageChange(model, language) {
     await wait.queue();
+    const previousAttributes = model.previousAttributes();
+    const previousLanguage = previousAttributes._activeLanguage;
     offlineStorage.set('lang', language);
     // set `_isStarted` back to `false` when changing language so that the learner's answers
     // to questions get restored in the new language when `_restoreStateOnLanguageChange: true`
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

js/data.js
```diff
@@ -169,7 +171,7 @@ class Data extends AdaptCollection {
       try {
         components.getModelName(modelData);
       } catch (error) {
-        logging.error(`Failed to load object ${modelData.__path__}${Object.prototype.hasOwnProperty.call(modelData, '__index__') ? `[${modelData.__index__}]` : ''}`)
+        logging.error(`Failed to load object ${modelData.__path__}${Object.prototype.hasOwnProperty.call(modelData, '__index__') ? `[${modelData.__index__}]` : ''}`);
```
danielghost, line 174:
```suggestion
        logging.error(`Failed to load object ${modelData.__path__}${Object.hasOwn(modelData, '__index__') ? `[${modelData.__index__}]` : ''}`);
```

oliverfoster: https://caniuse.com/mdn-javascript_builtins_object_hasown

danielghost (approved):

joe-allen-89 (approved):
👍

PR #344, opened 2023-04-04
Fix: ChildView API broken

oliverfoster:
fixes https://github.com/adaptlearning/adapt-contrib-branching/issues/43

### Fix
* Previous [pr](https://github.com/adaptlearning/adapt-contrib-core/pull/281) broke API where variable `model` doesn't exist.
https://github.com/adaptlearning/adapt-contrib-core/blob/3668a15c1e9777c23615746cd1d8465f9d729a27/js/views/adaptView.js#L299

### DIFF
--- a/js/views/adaptView.js
+++ b/js/views/adaptView.js
@@ -205,7 +205,7 @@ class AdaptView extends Backbone.View {
    */
   addChildView(childView) {
     const childModel = childView.model;
-    const type = childModel.get('_type')
+    const type = childModel.get('_type');
     const childViews = this.getChildViews() || [];
     childViews.push(childView);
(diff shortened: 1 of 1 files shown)


oliverfoster:
@sarveshwar-gavhane

cahirodoherty-learningpool (approved):
👀

oliverfoster:
> The change resolves the issue, but there is another unrelated branching issue when a branched assessment is reset. The following line errors when attempting to find a branched model id in the data: https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/a11y.js#L219

Can you move to a new issue and supply the error please?

danielghost (approved):

PR #351, opened 2023-04-14
Fix: Safari label click for input

### LINKED ISSUE #350: Safari onBlur onFocus onChange and click on label for input
### Description
In Safari, a click triggers a mousedown, a blur, a mouseup, a focus, then a click, which is different from chrome-based browsers. [Reference](https://sitr.us/2011/07/28/how-mobile-safari-emulates-mouse-events.html)

Sometimes we rerender the template on a blur, this rerender prevents the click event on the `label` from firing on the `input`, in Safari, when using JSX templates.

Instead of allowing the click to fall through from the `label` to the `input` as expected from the browser, we should prevent the `label` click event and trigger a manual click on the `input` referenced by the label's `for` attribute, side-stepping the rerender issue in Safari and producing the expected behaviour.

This fixes the issue in both the iOS and MacOS versions of Safari

References:
* https://www.eventbrite.com/engineering/a-story-of-a-react-re-rendering-bug/
* https://sitr.us/2011/07/28/how-mobile-safari-emulates-mouse-events.html
* https://stackoverflow.com/questions/9335325/blur-event-stops-click-event-from-working

Other attempts to fix:
* https://github.com/adaptlearning/adapt-contrib-mcq/pull/204
* https://github.com/adaptlearning/adapt-contrib-mcq/pull/200
* https://github.com/adaptlearning/adapt-contrib-mcq/pull/195
* https://github.com/adaptlearning/adapt-contrib-mcq/pull/173
* https://github.com/adaptlearning/adapt-contrib-mcq/pull/163

All other attempts at this so far, which include different `onChange`, `defaultCheck`, `onClick`,  `checked` combinations for the `input`, different `aria-label` and `label` positions and a variety of `preventDefault` behaviours on `onClick` and `mousedown` haven't worked in at least one of the various screen reader, browser or device combinations. Keyboard support means we need will always need to highlight an input `label` or container when an `input` element receives focus, which will invariably hit this Safari + JSX issue.

oliverfoster:
fixes #350 

### Fix
* General fix for safari label clicks to redirected the label click, to an input click, which overcomes complex brower+react bug

### Testing
Please test on ios and macos safari and on windows chrome, edge and firefox if you can.
Requires https://github.com/adaptlearning/adapt-contrib-mcq/pull/206

### DIFF
--- a/js/fixes.js
+++ b/js/fixes.js
@@ -1,2 +1,3 @@
 import 'core/js/fixes/img.lazyload';
 import 'core/js/fixes/reactHelpers.html';
+import 'core/js/fixes/safari.label.click.blur';
--- a/js/fixes/safari.label.click.blur.js
+++ b/js/fixes/safari.label.click.blur.js
@@ -17,7 +17,7 @@ import 'core/js/templates';
 
 Adapt.on('app:dataReady', () => {
   const config = Adapt.config.get('_fixes');
-  if (config?._jsxReactHelpersHTML === false) return;
+  if (config?._safariLabelClickBlur === false) return;
   applySafariLabelClickBlur();
 });
 
--- a/js/fixes/safari.label.click.blur.js
+++ b/js/fixes/safari.label.click.blur.js
@@ -1,0 +1,34 @@
+import Adapt from 'core/js/adapt';
+import 'core/js/templates';
+
+/**
+  In Safari, a click triggers a mousedown, a blur, a mouseup, a focus, then a click,
+  which is different from chrome-based browsers.
+  Sometimes we rerender the template on a blur, the rerender prevents
+  the click event from firing in Safari, instead of allowing click to fall through from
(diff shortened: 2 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

swashbuck:
This still isn't working for me 😭 

Mac OS Big Sur 11.7.6 - Safari 16.4.1
iOS 16.3.1 - Safari 16

oliverfoster:
It may be that return early statement I changed. I'll have a look.

oliverfoster:
Good catch. Thanks.

swashbuck (approved):
👍 Works on both Mac OS and iOS Safari

joe-allen-89 (approved):
Tested on windows with Chrome/Firefox/Edge all seem to be working as expected.

PR #352, opened 2023-04-18
Fix: _shouldResetAllAnswers moved into core from matching

oliverfoster:
fixes https://github.com/adaptlearning/adapt-contrib-matching/issues/171

Allows only incorrect answers to reset instead of all answers.

### Fix
* Added `_shouldResetAllAnswers ` as an `ItemsQuestionModel` settings

### DIFF
--- a/js/models/itemsQuestionModel.js
+++ b/js/models/itemsQuestionModel.js
@@ -192,8 +192,11 @@ export default class ItemsQuestionModel extends BlendedItemsComponentQuestionMod
   }
 
   resetItems() {
-    if (this.get('_shouldResetAllAnswers')) this.resetActiveItems();
-    else this.resetIncorrectItems();
+    if (this.get('_shouldResetAllAnswers') === false) {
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

joe-allen-89 (commented):
Tested with v5.31.1/Windows/Chrome, the reset of just the incorrect answers works nicely. 

@oliverfoster do you think that the correct answers should remain disabled on reset? If so not sure if we want to address that with this PR or a separate enhancement.

oliverfoster:
Is there a property to disable items on the model and are the views expecting that property? (I think matching is the only this that uses this behaviour at present). 

If the view isn't expecting a disabled property, then a separate pr should probably be made.

If the view is expecting a disabled property and it doesn't exist on the model then this pr should be updated accordingly.

StuartNicholls (approved):

PR #361, opened 2023-04-26
New: Added DOM Element Modification API

### LINKED ISSUE #360: Add API for listening to DOM element modifications
### Subject of the issue
Sometimes it's necessary to listen for dom modifications using the [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe) browser API. A standardised API can be provided in Adapt to make this API useful for handling many types of dynamically injected content in Adapt. By providing a single point of observing changes, we can reduce the processing overhead for this type of operation and stabalise its utilisation.

Current uses:
* [GraphicLottie](https://github.com/cgkineo/adapt-graphicLottie) Allows the use of animated svgs in place of img, it does this by replacing img tags with svg players
* [Navigation Button API](https://github.com/adaptlearning/adapt-contrib-core/pull/339) Listens for injected buttons in order to make models and views for managing them via an API

Possible future uses:
* [Tooltips API](https://github.com/adaptlearning/adapt-contrib-core/pull/207) To allow permanent tooltips to be rendered when a `[data-tooltip-id]` element is injected
* [Resources](https://github.com/adaptlearning/adapt-contrib-resources) To inject resources into content using a simple `<resources />` style tag
* [Glossary](https://github.com/adaptlearning/adapt-contrib-glossary) To inject clickable areas into content using a simple `<glossary term="Term" />` style tag

oliverfoster:
fixes #360 

### Fix
* Added DOM Element Modification API

### DIFF
--- a/js/DOMElementModifications.js
+++ b/js/DOMElementModifications.js
@@ -1,0 +1,255 @@
+import Backbone from 'backbone';
+import DOMElementModificationEventObject from './DOMElementModificationEventObject';
+
+/**
+ * Allows the capture of DOM descendant nodes on addition, removal or change.
+ * This is an Adapt orientated implementation of the browser MutationObserver API
+ * Reference: https://developer.mozilla.org/pt-BR/docs/Web/API/MutationObserver
+ *
+ * @example
+ * import documentModifications, { DOMElementModifications } from 'core/js/DOMElementModifications';
+ *
+ * documentModifications.on('added:[data-tooltip-id]', function onToolTipTargetAdded (event) {
+ *
+ * })
+ *
+ * this.listenTo(documentModifications, 'remove:[data-tooltip-id]', function onToolTipTargetRemoved (event) {
+ *
+ * });
+ *
+ * const navModifications = new DOMElementModifications({
+ *   el: document.querySelector('.nav__inner'),
+ *   watchAttributes: false,
+ *   watchImmediateChildrenOnly: true
+ * });
+ * navModifications.on('added removed', function onChildElementAddedOrRemoved (event) {
+ *
+ * })
+ *
+ * @event DOMElementModifications#added:selector Element matching selector has been added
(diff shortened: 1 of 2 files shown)


nachocinalli:
I can't listen for the 'added:img' event within the notify popup.
I have noticed that it works for example in Hotgrid 4.3.1 (with .hbs templates), but not in higher versions where JSX is used.

oliverfoster:
It seems to be that the element is added and then immediately changed by react, which identifies a flaw in my algorithm that registers this as a "changed" event only. I will fix and update.

oliverfoster:
That should fix it hopefully.
image

nachocinalli:
works, thank you.

js/DOMElementModifications.js
```diff
@@ -0,0 +1,287 @@
...
+        const EventObject = new DOMElementModificationEventObject({
+          type: eventName,
+          target,
+          changedAttributes,
+          previousAttributes
+        });
+        selectors.forEach(selector => {
+          if (selector && selector !== true) {
+            // Event attachment has specified selector
+            if (eventName !== 'changed') this.trigger(`${eventName}:${selector}`, EventObject);
```
cahirodoherty-learningpool, line 192: I think the if check can be removed and line 193 deleted as they both achieve the same thing

oliverfoster: 192 does the `added`/`removed` events (when an explicit listen has been attached), 193 does the `changed` events, either subsequent to the `added`/`removed one, or on its own, always.

cahirodoherty-learningpool: ah good point. My mistake, sorry. I had assumed a `return` that wasn't actually there 👍

cahirodoherty-learningpool (approved):
👀

StuartNicholls (approved):

PR #366, opened 2023-04-30
New: Introduce Menus as Pages

### LINKED ISSUE #365: Allow articles blocks and components on menus
### Subject of the issue
By treating menus as pages, but pages with child content objects as well as child articles, we can make more interesting menus by using components to create them.

oliverfoster:
fixes #365 

With `config.json:_isPageMenu = true`, menus may have articles, blocks and components, menus are rendered as pages. Use `ComponentMenuView` and `ComponentMenuModel` as the basis for menu components, which will inherit the children and therefore completion from the nearest contentobject and render the contentobject's children by default.

### New
* Added `ComponentMenuView` to render child views by default
* Added `ComponentMenuModel` to inherit the children and therefore completion, from its immediate `contentobject` ancestor
* Added `config.json:_isPageMenu` to register `PageView` for `course` and `menu` `_type`.
* Restricted `AdaptView` so that a default page doesn't render child `contentobject` types
* Updated `getAllDescendantModels` to take a `filter` argument as part of an `options` object

### Testing
* Install https://github.com/cgkineo/adapt-contrib-menuBox
* Remove default menu
* Add `config.json:_isPageMenu = true`
* Add `course.json:_pageLevelProgress._showAtCourseLevel = true`
* Add an article with a `_parentId = "course"`, along with a block and a component, which should have `_component = "menuBox"`
* Add any other articles, blocks and components to the course page

Or use the following zip, extract and run `npm install` then build as usual:
[test.zip](https://github.com/adaptlearning/adapt-contrib-core/files/11362330/test.zip)

### Todo
* Bring to AATv1 - allow menus to be edited as pages when `_isPageMenu = true`, default to `_isPageMenu` when no menu is selected @taylortom 
* Remove page header for a banner component https://github.com/adaptlearning/adapt_framework/issues/2703

pagemenus

### DIFF
--- a/js/models/ComponentMenuModel.js
+++ b/js/models/ComponentMenuModel.js
@@ -1,0 +1,35 @@
+import ComponentModel from 'core/js/models/componentModel';
+import data from '../data';
+
+class ComponentMenuModel extends ComponentModel {
+
+  setupModel() {
+    this.setupChildListeners();
+    this.init();
+    _.defer(() => {
+      this.checkCompletionStatus();
+      this.checkInteractionCompletionStatus();
+      this.checkLocking();
+      this.checkVisitedStatus();
+      this.setupTrackables();
+    });
+  }
+
+  getChildren() {
+    if (this._childrenCollection) {
+      return this._childrenCollection;
+    }
+
+    const parentContentObject = this.findAncestor('contentobject');
+    const id = parentContentObject.get('_id');
+    // Look up child by _parentId from data
+    const children = data.filter(model => model.get('_parentId') === id && model.isTypeGroup('contentobject'));
+    const childrenCollection = new Backbone.Collection(children);
+
+    this.setChildren(childrenCollection);
+    return this._childrenCollection;
+  }
+
+}
+
+export default ComponentMenuModel;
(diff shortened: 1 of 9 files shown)


StuartNicholls:
I have actually started work on a page header / hero plug-in, will add some layouts we've mapped out on the issue #2703

oliverfoster:
Aria-levels are off by one for components and group sub items.

oliverfoster:
The confusing part about this is that the children of the Page Menu don't run in render order.

Ideally:
| Parent Name | Child name |
| -- | -- |
| Page Menu | Article 1 |
| | SubPage 1 |
| | Article 2 |
| | SubPage 2 |
| | Article 3 |
| | SubPage 3 |

Whereas because articles live in `article.json` and pages live in `contentObjects.json` and the Page Menu can live in `course.json`, we have the children of the Page Menu children split over two separate files and their order isn't reflective of the render order.

| course.json | contentObject.json | articles.json |
| -- | -- | -- | 
| Page Menu | SubPage 1 | Article 1 |
| |SubPage 2 | Article 2 |
| | SubPage 3 | Article 3 |

PR #368, opened 2023-05-03
Fix: Scroll background from popups

### LINKED ISSUE #367: Able to scroll the background content in modal view
**Identified issue:**

-  Able to scroll the background content when accessing the overlay area.
-  When scrolling the content within the modal (using the mouse wheel), and the scrollbar comes to the end of the content (top or bottom), then if you keep on scrolling the background content starts to scroll.  

**Expected behaviour:**
When a modal window are open, it should block the accessibility of the background content.

Because of this new update in https://github.com/adaptlearning/adapt-contrib-core/commit/18391922a770faeea65564d5bcd0a3170cf77ecd

### Your environment
* v5.30.3
* Chrome and firefox
* Window 10

### Screenshots (if you can)
Hot Graphic-modal view out of bounds scroll

oliverfoster:
fixes #367 

Tested on Chrome Android and Windows with mousewheel, touch and keyboard interactions.

### Fix
* Fixed typo in previous commit https://github.com/adaptlearning/adapt-contrib-core/commit/18391922a770faeea65564d5bcd0a3170cf77ecd
* In order to sidestep Safari 16.4 issues, DOM event attachment is permanent and returned early rather than toggled attached and removed

### DIFF
--- a/js/a11y/scroll.js
+++ b/js/a11y/scroll.js
@@ -183,7 +183,7 @@ export default class Scroll extends Backbone.Controller {
     const isTouchEvent = event.type === 'touchmove';
     const isKeyDownEvent = event.type === 'keydown';
     const hasTouchStartEvent = this._touchStartEventObject?.originalEvent;
-    if ((isTouchEvent && !hasTouchStartEvent) || !isKeyDownEvent) {
+    if ((isTouchEvent && !hasTouchStartEvent) && !isKeyDownEvent) {
       return $target;
     }
     const directionY = this._getScrollDirection(event);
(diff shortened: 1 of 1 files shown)


kirsty-hames (commented):
I could replicate the issue on Mac Safari/FF/Chrome and confirm fixed for FF/Chrome. It's still an issue for Safari.

oliverfoster:
Can't repliace on MacOS Safari Version 15.5 (17613.2.7.1.8) locally, can see bug on screenshare with Kirsty on Version 16.4 (18615.1.26.110.1) using both the touchpad and a mouse. Cannot yet find a reference to a Safari bug. It seems that the wheel event handlers aren't firing. Upgrading my Safari to see if I can replicate.

oliverfoster:
The event attach/detach process seems to cause issues on Safari 16.2. I have switched to keeping the listeners always on and returning early in the handlers instead.

kirsty-hames (approved):
That's fixed thanks @oliverfoster. Retested on Mac Safari, FF and Chrome.

joe-allen-89 (approved):
👍 Tested on Windows with Chrome/Firefox/Edge, all working nicely.

PR #394, opened 2023-06-26
Fix: Added _isRendered: true to main view render

### LINKED ISSUE #393: Fix: Main view is never marked as _isRendered
### Subject of the issue
This pr https://github.com/adaptlearning/adapt-contrib-core/pull/388:
https://github.com/adaptlearning/adapt-contrib-core/blob/a6304f4cbb16d5c827a4ebbc9afb253f0898ead1/js/views/adaptView.js#L87-L106
Adds a check when rendering, to see if the model has been marked as `_isRendered` from the `addChildren` function.
https://github.com/adaptlearning/adapt-contrib-core/blob/6ef84972f04cb760fb2862b0ba44990aeca73a24/js/views/adaptView.js#L149-L197
Which means that main menus made from jsx won't render correctly, as the router never sets the main view is `_isRendered: true`.

The lifecycle of the `_isRendered` property should be set to `true` when the view is being rendered by the router or by the view's `addChildren` property and then removed when `Adapt.remove()` has been called.
https://github.com/adaptlearning/adapt-contrib-core/blob/6ef84972f04cb760fb2862b0ba44990aeca73a24/js/adapt.js#L196-L212

oliverfoster:
fix #393 

### Fix
* Added _isRendered: true to the main view render

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -230,7 +230,7 @@ export default class AdaptModel extends LockingModel {
     if (!canReset) return false;
     const isHardReset = (type === 'hard' || type === true);
     const isSoftReset = (type === 'soft');
-    if (!isHardReset && !isSoftReset) return false
+    if (!isHardReset && !isSoftReset) return false;
     const resetData = {
       _isEnabled: true,
       _isInteractionComplete: false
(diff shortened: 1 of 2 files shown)


guywillis (approved):
👀

joe-allen-89 (approved):
👍

cahirodoherty-learningpool (approved):
👀

PR #398, opened 2023-06-30
New: Added new _srcFocalPoint class to image.jsx (fixes #396)

### LINKED ISSUE #396: image.jsx template - pass through new `_srcFocalPoint` property
Allow a new property, `_srcFocalPoint`, to be passed through from plugins to the image.jsx template.

The new property will allow end users the ability to change the focal point of the image from the plugin.  

### Your environment
* Master

Example
```
"_graphic": {
  "_src": "course/en/images/single-width.png",
  "_srcFocalPoint": "left"
},

<img class="component__image graphic__image object-position-left" src="course/en/images/single-width.png" aria-label="" aria-hidden="true" loading="eager">
```

guywillis:
Resolves: #396 

### New
* Added a new class to image template, to be based on `_srcFocalPoint` value that's part of the `_graphic` object

### DIFF
--- a/templates/image.jsx
+++ b/templates/image.jsx
@@ -34,7 +34,10 @@ export default function Image(props) {
     >
 
       <img
-        className={prefixClasses(props.classNamePrefixes, ['__image'])}
+        className={classes([
+          prefixClasses(props.classNamePrefixes, ['__image']),
+          props?._srcFocalPoint && `object-position-${props?._srcFocalPoint}`
+        ])}
         src={src}
         aria-label={a11y.normalize(props.alt)}
         aria-hidden={!props.alt}


templates/image.jsx
```diff
@@ -34,7 +34,10 @@ export default function Image(props) {
     >
 
       <img
-        className={prefixClasses(props.classNamePrefixes, ['__image'])}
+        className={classes([
+          prefixClasses(props.classNamePrefixes, ['__image']),
+          props?._srcFocalPoint && `object-position-${props?._srcFocalPoint}`
```
cahirodoherty-learningpool, line 39: Hi Guy, this seems like a very specific class name to apply, but I can't see anywhere in Core or Vanilla that currently picks it up. Are you intending to use it to apply the [CSS object-position](https://developer.mozilla.org/en-US/docs/Web/CSS/object-position) setting somewhere automatically?

guywillis: Hey Cahir, you're correct it's not currently being used by any plugins. The intention is to bake in the potential for changing the images object position going forwards should we require it.

kirsty-hames: I think the ability to set the `object-position` is a useful additional. I think implementing this as a `class` though could be limiting.
1, the classes need to be predefined,
2, typically you cover keyword values as classes (`top`, `left`, `bottom`, `right`) ruling out percentage, length and off set values. Even with those keywords I tend to double up e.g. `right bottom`.

For Vanilla background image `background-position`, we set [CSS styles](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/59e377d7a2da38ab29c596a278b7ce8374dfbe52/js/themeView.js#L70) rather than classes. What are thoughts on doing the same here?

swashbuck: My two cents... We could create classes for all of the keyword values. For instance:

`object-position-top-left`
`object-position-top-right`
`object-position-bottom-left`
`object-position-bottom-right`
etc....

Then, custom theme work can be utilized for more granular control over the positioning (ex. `left 4% bottom 2%`). However, I think the keyword options would fit the majority of use cases.

swashbuck, line 39:
```suggestion
          props._srcFocalPoint && `object-position-${props._srcFocalPoint}`
```
@guywillis We shouldn't need the optional chaining (?.) here since React will just ignore it if the property is undefined. No risk of throwing an error.

Also, wondering if this property should drop the underscore and just use `srcFocalPoint`? Other properties that get passed to this template like `classNamePrefixes` are also not translatable and don't use an underscore. So, that would make it consistent.

oliverfoster: `_srcFocalPoint` will often be supplied straight from the json,  where it would require the `_` as it is not translatable, removing the `_` from here would require code to remove the underscore from the json attribute.

`classNamePrefixes` is a code supplied property, usually of the component utilising the behaviour, `hotgraphic`, `accordion` etc, not a json supplied property.

oliverfoster (approved):

joe-allen-89 (approved):

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

PR #403, opened 2023-07-06
Fix: Add draggable attribute option (fixes #402)

### LINKED ISSUE #402: Add draggable attribute to image.jsx template
### Subject of the issue
In https://github.com/adaptlearning/adapt-contrib-narrative/pull/250, the non-standard `user-drag` CSS property was replaced with the standardized `draggable` HTML property. Currently, _image.jsx_ does not seem to support this option.

We need an option to include `draggable="false"` on images that utilize _image.jsx_.

Note that this should not use the shorthand method ``. The correct usage is ``

### References
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable
https://caniuse.com/webkit-user-drag

### Your environment
* AF 5.31.12

swashbuck:
Fixes #402

### Fix
* Adds option for the HTML attribute `draggable`

### Testing
1. Where the image.jsx template is used, add the draggable property. For example:

```
<templates.image {..._graphic}
  draggable="false"
/>
```
2. If `draggable` is set, it should include the property on the `` tag.
3. If `draggable` is not set, it should _not_ include the property on the `` tag.

This will be used in https://github.com/adaptlearning/adapt-contrib-narrative/pull/270, so it can also be tested there.

### DIFF
--- a/templates/image.jsx
+++ b/templates/image.jsx
@@ -40,7 +40,7 @@ export default function Image(props) {
         aria-hidden={!props.alt}
         loading='eager'
         aria-describedby={props.longdescription ? props.longDescriptionId : undefined}
-        draggable={props.draggable || null}
+        draggable={props.draggable ?? null}
       />
 
       {props.attribution &&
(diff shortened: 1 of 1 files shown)


kirsty-hames (approved):
Tested with https://github.com/adaptlearning/adapt-contrib-narrative/pull/270 on Mac Safari, Chrome and iPhone

cahirodoherty-learningpool (approved):
👀

joe-replin (approved):
Approved.

PR #404, opened 2023-07-07
Fix: Create focusAfterReset to move focus behavior from onResetClicked (fix #391)

### LINKED ISSUE #391: Question View - Move focus action outside of onResetClicked()
When the reset button is clicked on a component, onResetClicked() is called in questionView.js. As part of this function, the focus is moved to the first readable item in the component.

https://github.com/adaptlearning/adapt-contrib-core/blob/6ef84972f04cb760fb2862b0ba44990aeca73a24/js/views/questionView.js#L304

If a question component wants to change the focus behavior (ex. having it focus on an interactive item instead of the component heading), this entire function must be overridden to change the above line.

### Suggestion
Move the focus line to a new function so that question components can easily override it.

### Your environment
* Core 6.36.2

swashbuck:
Fixes #391 

### Fix
* Create `focusAfterReset()` which moves the focus behavior from `onResetClicked()`. This makes it simpler to override the focus behavior in a plugin.

### DIFF
--- a/js/views/questionView.js
+++ b/js/views/questionView.js
@@ -311,7 +311,7 @@ class QuestionView extends ComponentView {
 
   focusAfterReset() {
     // Focus on the first readable item in this element
-    a11y.focusNext(this.$el, { preventScroll: true });
+    a11y.focusNext(this.$el);
   }
 
   /**
--- a/js/views/questionView.js
+++ b/js/views/questionView.js
@@ -300,14 +300,18 @@ class QuestionView extends ComponentView {
     const currentModel = data.findById(location._currentId);
     // Make sure the page is ready
     if (!currentModel?.get('_isReady')) return;
-    // Focus on the first readable item in this element
-    a11y.focusNext(this.$el, { preventScroll: true });
+    this.focusAfterReset();
   }
 
   setQuestionAsReset() {
     this.model.setQuestionAsReset();
     this.resetQuestion();
     this.$('.component__widget').removeClass('is-submitted');
+  }
+
+  focusAfterReset() {
+    // Focus on the first readable item in this element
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

js/views/questionView.js
```diff
@@ -310,6 +309,11 @@ class QuestionView extends ComponentView {
     this.$('.component__widget').removeClass('is-submitted');
   }
 
+  focusAfterReset() {
+    // Focus on the first readable item in this element
+    a11y.focusNext(this.$el, { preventScroll: true });
```
oliverfoster, line 314: I'm not sure you want this: `{ preventScroll: true }`

swashbuck: Removed. I agree that it would make sense to scroll to the point where the new focus is. Otherwise, it could be off screen and the user could lose their place.

Do you know why this was there in the first place? Was it to avoid taking control away from users who use assistive technologies? Is there an a11y concern about controlling focus when the user prefers reduced motion perhaps? Can't seem to find any a11y guidelines about this.

oliverfoster: All of the preventScroll behaviours are just to prevent jumping for sighted users. But in this case, I don't think it makes sense. We'll see.

> Was it to avoid taking control away from users who use assistive technologies? 

No, just to prevent jumping.
 
> Is there an a11y concern about controlling focus when the user prefers reduced motion (ex. smooth scrolling could still pose an issue for these users)? 

No, reduced motion is predominantly about aesthetic animations not functional jumps, I believe.

oliverfoster (approved):

kirsty-hames (approved):
👀

PR #406, opened 2023-07-11
Fix: Return null in getChildViews if this._childViews is undefined (fix #405)

### LINKED ISSUE #405: getChildViews() - Explicitly return null if _childViews is undefined
https://github.com/adaptlearning/adapt-contrib-core/blob/1e73f329c21e6eb906c102db23f600666117cd74/js/views/adaptView.js#L397

This should return `null` when undefined instead of returning the "undefined" `this._childViews` value. Should probably be:

```
if (!this._childViews) return null;
```

This would not change the behavior but will make the code's intentions clearer.

swashbuck:
Fixes #405 

### Fix
* Return `null` in `getChildViews()` if `this._childViews` is undefined. Minor fix for code readability.

### DIFF
--- a/js/views/adaptView.js
+++ b/js/views/adaptView.js
@@ -394,7 +394,7 @@ class AdaptView extends Backbone.View {
    * @returns {[AdaptView]}
    */
   getChildViews() {
-    if (!this._childViews) return this._childViews;
+    if (!this._childViews) return null;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

eleanor-heath (approved):

PR #435, opened 2023-08-28
New: Add html classes depending on the loading screen's visibility (fixes #423)

### LINKED ISSUE #423: Add html class after all assets are loaded for a page
To help with the timing of CSS animations, it would be useful to have an `html` class that indicates when all assets have been loaded for a page.

### Use case
I want to animate the title element as soon as the page is loaded and the loading screen is removed. There is currently not a way to know when the loading screen is removed, so it's difficult to know when the animation should be triggered. For instance, there is a chance that the animation is triggered while the loading screen is still overlapping the content, so the animation will never be seen. With a new class like `page-view-ready`, I can do the following:

```
.page__title {
  transition: 0.5s all ease-in-out;
}

html.page-view-ready .page__title {
  color: fuchsia;
  transform: scale(1.05) rotate(-1deg) translate(1rem, -1rem);
}
```

https://github.com/adaptlearning/adapt-contrib-vanilla/assets/898168/4b36dc74-f705-4b71-8268-607a1be1ee04

swashbuck:
Fixes #423 

### Fix
* TBD

### Testing
1. Steps for testing

### Related
Language Picker and Spoor should be updated to use `router.hideLoading()` instead of the current `$('.js-loading').hide()`.

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -41,10 +41,12 @@ class Router extends Backbone.Router {
   }
 
   showLoading() {
+    $('html').removeClass('is-loading-hidden').addClass('is-loading-visible');
     $('.js-loading').show();
   }
 
   hideLoading() {
+    $('html').addClass('is-loading-hidden').removeClass('is-loading-visible');
     $('.js-loading').hide();
   }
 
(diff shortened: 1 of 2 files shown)


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):

PR #437, opened 2023-08-31
Fix: ENUM refactor, linting and fix aliases (fixes #436)

oliverfoster:
fixes #436 

### Fix
* Refactor to ES6
* Fix alias enumeration behaviour
* Added `toString()` which is more standard than `asString`
* Exposed `lookupHash` for easy debugging

PR #438, opened 2023-09-01
Update: Added NOT ATTEMPTED enum to COMPLETION_STATE (fixes #424)

### LINKED ISSUE #424: Expand COMPLETION_STATE enums
Expand https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/enums/completionStateEnum.js to account for the different values as per https://github.com/adaptlearning/adapt-contrib-spoor/pull/280#discussion_r1279112556?

Do we want all the values in here though as you could argue that some are specific to SCORM? Outside of spoor, no other hardcoded instances occur within the framework. https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/tracking.js only uses the currently available enums.

Which additional values would we want:
- [x] "completed"
- [x] "passed"
- [x] "failed"
- [x] "incomplete"
- [ ] "browsed"
- [x] "not attempted"
- [ ] "not_attempted" (mentioned in SCORM 2004 spec - mapped to "not attempted")
- [ ] "unknown"

danielghost:
Fixes #424.

### Update
* Added `COMPLETION_STATE('NOT ATTEMPTED')` enum and its `COMPLETION_STATE.NOTATTEMPTED` alias.

### Dependency
* https://github.com/adaptlearning/adapt-contrib-core/pull/437


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #455, opened 2023-10-17
Fix: Boolean _isAtLeastOneCorrectSelection (fixes #454)

### LINKED ISSUE #454: `itemsQuestionModel._isAtLeastOneCorrectSelection` not a boolean
Since https://github.com/adaptlearning/adapt-contrib-core/commit/f9de89394ad418c9b7fdefbbd12b9659a090f827, `_isAtLeastOneCorrectSelection` returns a number rather than a boolean - https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/models/itemsQuestionModel.js#L95. This doesn't actually cause an issue, as any value > 0 will return true when used conditionally, but technically this isn't the expected value.

oliverfoster:
fixes #454 

### Fix
* Coerce `_isAtLeastOneCorrectSelection` to boolean

### DIFF
--- a/js/models/itemsQuestionModel.js
+++ b/js/models/itemsQuestionModel.js
@@ -92,7 +92,7 @@ export default class ItemsQuestionModel extends BlendedItemsComponentQuestionMod
 
     activeChildren.forEach(itemModel => itemModel.set('_isCorrect', itemModel.get('_shouldBeSelected')));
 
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

danielghost (approved):

joe-allen-89 (approved):
👍

Issue #463, opened 2023-10-31
Tooltips - Close tooltip before navigating to a new location


swashbuck:
### Subject of the issue
Before navigating to a new page, a button's tooltip should be removed. Currently, they're only being removed on blur/mouseleave. This is causing the tooltips to sometimes [persist](https://github.com/cgkineo/adapt-pageNav/pull/49) upon loading a new page.

### Steps to reproduce
1. Install Page Nav with this PR: https://github.com/cgkineo/adapt-pageNav/pull/49
2. Add a Page Nav component instance and enable tooltips.
3. In Safari, hover over a button (e.g. Previous or Root) and then click it.

### Expected result
Tooltip should not persist when the new location has finished loading.

### Actual behaviour
Tooltip persists and sometimes affects the page rendering

### Screenshots
From Page Nav PR [49](https://github.com/cgkineo/adapt-pageNav/pull/49#issuecomment-1768730191):
![root_1](https://github.com/cgkineo/adapt-pageNav/assets/7045330/0edfffc1-42e9-4bb2-a634-46853068c77d)

PR #464, opened 2023-10-31
Fix: Duplicate and persistent tooltip issues (fixes #429 #463)

### LINKED ISSUE #429: Tooltips duplicating on Firefox
When hovering over a nav item in Firefox a new tooltip is constantly added to the dom until you move your cursor away from the button. 

### Your environment
* v5.31.21 - core 6.39.0
* Firefox

### Steps to reproduce
Set up a course with tooltips on nav items, open in Firefox and hover over a nav button

### Expected behaviour
One tooltip is shown

### Actual behaviour
Repeatedly adds an additional tooltip to the dom.

swashbuck:
Fix #429 #463 

### Fix
* Fixes the issue where tooltips were being repeatedly created. Removes the unnecessary debounce on onMouseOver.
* Ensure tooltip is hidden when the element (e.g. a button) is clicked

### Testing (#429)
1. To test mouse functionality, hover over a navigation button that has tooltips enabled.
2. To test keyboard navigation, navigate to the button via keyboard.

### Expected result (#429)
The tooltip should only appear once and then disappear once the button loses focus or fires a mouseout event.

### Testing (#463)
1. Install Page Nav with this PR: https://github.com/cgkineo/adapt-pageNav/pull/49
2. Add a Page Nav component instance and enable tooltips.
3. In Safari, hover over a button (e.g. Previous or Root) and then click it.

### Expected result (#463 )
Tooltip should not persist when the new location has finished loading.

### DIFF
--- a/js/views/TooltipItemView.js
+++ b/js/views/TooltipItemView.js
@@ -129,6 +129,11 @@ export default class TooltipItemView extends Backbone.View {
     this.remove();
   }
 
+  onClick() {
+    console.log('onClick!!!');
+    this.remove();
+  }
+
   onMouseOut() {
     this.remove();
   }
--- a/js/views/TooltipView.js
+++ b/js/views/TooltipView.js
@@ -18,7 +18,7 @@ export default class TooltipView extends Backbone.View {
   }
 
   initialize() {
-    _.bindAll(this, 'onMouseOver', 'onKeyDown', 'onMouseOut');
+    _.bindAll(this, 'onMouseOver', 'onKeyDown');
     this._tooltipData = {};
     this._tooltips = [];
     this.listenToOnce(Adapt, 'adapt:preInitialize', this.onAdaptPreInitialize);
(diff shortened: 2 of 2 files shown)


js/views/TooltipItemView.js
```diff
@@ -129,6 +129,11 @@ export default class TooltipItemView extends Backbone.View {
     this.remove();
   }
 
+  onClick() {
```
oliverfoster, line 132: I can't see where this function is being called from. There's no click event listener in this view.

oliverfoster (commented):
It had always thought that it is usual and expected in the UX world that there is a delay between the mouseover and tooltip being shown.

oliverfoster:
Please try this pr https://github.com/adaptlearning/adapt-contrib-core/pull/465 @swashbuck

js/views/TooltipItemView.js
swashbuck: It's called on line 36:

`$(document).on('click', '[data-tooltip-id]', this.onClick);`

oliverfoster: Different file.

swashbuck: Dah, you're right. Must have still been there from an earlier attempt to fix it.

PR #467, opened 2023-11-02
Fix: incorrect `scrolledTo` event name (fixes #399).

### LINKED ISSUE #399: incorrect `scrolledTo` event name
The `scrolledTo` event trigger at https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/router.js#L423, uses the `Location` class object in its name rather than the same `newLocation` value sent for the `scrollTo` event. This currently prevents the ability to listen to this event.

Amend line to:
```JavaScript
Adapt.trigger(`${newLocation}:scrolledTo`, selector);
```

danielghost:
Fixes #399.

### Fix
* Incorrect `scrolledTo` event name which was preventing this from being used by listeners.

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -426,7 +426,7 @@ class Router extends Backbone.Router {
     await new Promise(resolve => {
       _.delay(() => {
         a11y.focusNext(selector);
-        Adapt.trigger(`${location}:scrolledTo`, selector);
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):

joe-allen-89 (approved):
👍

PR #469, opened 2023-11-14
Update: added _lang property for AAT multilang support (fixes #468)

### LINKED ISSUE #468: Add _lang property

chris-steele:
#468 

### Update
* Added support for multilang in AAT v1

### DIFF
--- a/schema/content.schema.json
+++ b/schema/content.schema.json
@@ -10,7 +10,10 @@
     "_friendlyId": {
       "type": "string",
       "title": "Friendly id",
-      "description": "A human-readable ID for this content object which will replace the standard _id field on course build (useful in instances where the _id must be specified in the config of an extension). Must be unique to this course."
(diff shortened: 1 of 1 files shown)


oliverfoster:
Should this go in both schema versions?

schema/content.schema.json
```diff
@@ -25,6 +28,15 @@
         "editorOnly": true
       }
     },
+    "_lang": {
+      "type": "string",
+      "title": "Language",
+      "description": "The language/locale e.g. 'en' or 'fr-CA'.",
+      "default": "en",
+      "_backboneForms": {
+        "editorAttrs": { "disabled": true }
```
taylortom, line 37: You can also set `_backboneForms.showInUi` to `false` to hide the input completely (not sure which behaviour would be most appropriate).

chris-steele: good to know, yep possibly better hidden as lang menu shows currently selected language and user will never be able to edit this field

chris-steele:
> Should this go in both schema versions?

@oliverfoster I don't think so as we're not backporting multilang to legacy

PR #473, opened 2023-11-21
Fix: Detached elements are not focusable or readable (fixes #472)

### LINKED ISSUE #472: a11y.isFocusable returns true on detached elements
### Subject of the issue
https://github.com/adaptlearning/adapt-contrib-core/blob/805e85bd87be176b719f2418a90fea795adb4c94/js/a11y.js#L436-L453

oliverfoster:
fixes #472 

### Fix
* a11y.isReadable and therefore a11y.isFocusable returns false if element is not in the body tag

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -448,7 +448,8 @@ class A11y extends Backbone.Controller {
         style.visibility === 'hidden' ||
         item.getAttribute('aria-hidden') === 'true';
     });
-    if (isNotVisible) {
+    const isInDOM = Boolean($element.parents('body').length);
+    if (isNotVisible || !isInDOM) {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


js/a11y.js
```diff
@@ -448,7 +448,8 @@ class A11y extends Backbone.Controller {
         style.visibility === 'hidden' ||
         item.getAttribute('aria-hidden') === 'true';
     });
-    if (isNotVisible) {
+    const isInDOM = Boolean($element.parents('body').length);
```
cahirodoherty-learningpool, line 451: Minor: Could this `isInDOM` check be done and early false return before the isNotVisible computation?

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #477, opened 2023-11-30
Update: Prevent Notify popup rendering empty .notify__section container (fixes #476)

### LINKED ISSUE #476: Prevent popup from rendering empty .notify__section
### Subject of the issue
The `.notify__section` was implemented during a recent [PR](https://github.com/adaptlearning/adapt-contrib-core/pull/223) to expand feedback and add image support. The `.notify__section` element is rendered regardless if the Notify body or graphic is used. For example, plugins that create their own Notify view such as Hotgraphic.

empty_notify_section

### Expected behaviour
Don't render `.notify__section` unless `_graphic.src` and/or `body` exists.

Issue initially raised here for reference https://github.com/cgkineo/adapt-hint/issues/16

kirsty-hames:
Don't render `.notify__section` unless `_graphic.src` and/or `body` exists.

Fixes https://github.com/adaptlearning/adapt-contrib-core/issues/476

**Question**
Just something I noticed during this PR. Does anyone know what the `~` is used for in the [notifyPopup.hbs](https://github.com/adaptlearning/adapt-contrib-core/blob/535905f91b1dfca190179f6e6ef2909e020f0efb/templates/notifyPopup.hbs#L56) conditions? 
In core, this is the only hbs file using it.

### DIFF
--- a/templates/notifyPopup.hbs
+++ b/templates/notifyPopup.hbs
@@ -26,7 +26,8 @@
             </div>
           </div>
         {{/if}}
-        
+
+        {{#any _graphic._src body}}
         <div class="notify__section">
           <div class="notify__section-inner">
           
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

PR #479, opened 2023-12-05
Fix: Allow classNamePrefixSeparator in image.jsx (fixes #478)

### LINKED ISSUE #478: Use of image.jsx when rendering items has bad classNamePrefixes
### Subject of the issue
This pr into hotgraphic https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/290 currently has a discrepency between the classnames in the original template and the classnames if/when using the image.jsx partial.

> Currently it comes out as .hotgraphic-popup__image-container from image.jsx rather than .hotgraphic-popup__item-image-container as it is in the styling.

It should be possible to change the classNamePrefix separator `__` in order to accomodate.

oliverfoster:
fixes #478 

### Fix
* Add option to override default className separator `__` in image.jsx

### DIFF
--- a/templates/image.jsx
+++ b/templates/image.jsx
@@ -22,12 +22,13 @@ export default function Image(props) {
   );
   const hasSource = Boolean(src);
   if (!hasSource) return null;
+  const classNamePrefixSeparator = (props.classNamePrefixSeparator || '__');
   const attributionClassNamePrefixes = (props.attributionClassNamePrefixes || props.classNamePrefixes);
   return (
     <span
       id={props.id}
       className={classes([
-        prefixClasses(props.classNamePrefixes, ['__image-container']),
+        prefixClasses(props.classNamePrefixes, [`${classNamePrefixSeparator}image-container`]),
         props.classes,
         props.attribution && 'has-attribution'
       ])}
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

kirsty-hames (changes requested):
`classNamePrefixSeparator` is applied to both `classNamePrefixes` and `attributionClassNamePrefixes`, however the image attribution doesn't require a prefix separator. 
Looking at the current naming in Adapt, regardless if the attribution is for a component image or component item image, we just specify the component and component type, e.g. `.component__attribution .accordion__attribution`. [Vanilla attribution styles](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/cb5fc98607b042487566bab99daf4c784f5f2afd/less/core/attribution.less#L6) target these selectors too.

I'd suggest we apply `classNamePrefixSeparator` to `classNamePrefixes` only. Would you like me to make the amend @oliverfoster whilst I'm working on https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/290 ?

oliverfoster:
Yes please. Take it away. :+1:

joe-allen-89 (approved):
👍

PR #487, opened 2024-01-24
Fix: Auto focus next only when returning to body from disabled or hidden (fixes #486)

### LINKED ISSUE #486: BrowserFocus auto next errors when intentionally moving focus from hidden element
### Subject of the issue
If [focusNext](https://github.com/adaptlearning/adapt-contrib-core/blob/69044466207afba515edf28c6f5e0a2eac566cbc/js/a11y.js#L708-L713) or similar is called and moves focus from an element which is intentionally offscreen, then this [if statement](https://github.com/adaptlearning/adapt-contrib-core/blob/69044466207afba515edf28c6f5e0a2eac566cbc/js/a11y/browserFocus.js#L50-L52) will cause adapt to autonext unintentionally [here](https://github.com/adaptlearning/adapt-contrib-core/blob/69044466207afba515edf28c6f5e0a2eac566cbc/js/a11y/browserFocus.js#L54).

Autonext on blur should only occur when the event related target (the element receiving focus), is the body tag and is receiving focus because the target element has been disabled or otherwise hidden.

oliverfoster:
fixes #486 

### Fix
* Stop auto next focus unless focus is returning to the body from a disabled or hidden element

### DIFF
--- a/js/a11y/browserFocus.js
+++ b/js/a11y/browserFocus.js
@@ -45,6 +45,10 @@ export default class BrowserFocus extends Backbone.Controller {
     if (event.target !== event.currentTarget) {
       return;
     }
+    // Do not auto next if the focus isn't returning to the body
+    if (!$(event.relatedTarget).is('body')) {
+      return;
+    }
     // Check if element losing focus is losing focus
     // due to the addition of a disabled class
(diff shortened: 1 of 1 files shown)


guywillis (approved):
👍

cahirodoherty-learningpool (approved):
👀

lemmyadams (approved):

PR #492, opened 2024-01-31
Fix: Restore user answers on setAttemptObject (fixes #491)

### LINKED ISSUE #491: componentModel.setAttemptObject does not restore user selections
### Subject of the issue
When getting the attemptObject from a clone of a model and then updating the original with the attempt object, I would expect the user selections to be modified accordingly.

The following code does not update the original model with the cloned model selections:
```js
originalModel.setAttemptObject(clonedModel.getAttemptObject());
```

oliverfoster:
fixes #491 

### Fix
* Restores user answers on setAttemptObject

### DIFF
--- a/js/models/componentModel.js
+++ b/js/models/componentModel.js
@@ -161,6 +161,7 @@ class ComponentModel extends AdaptModel {
    */
   setAttemptObject(object, silent = true) {
     this.set(object, { silent });
+    this.restoreUserAnswers();
   }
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

lemmyadams (approved):

joe-allen-89 (approved):
👍

PR #494, opened 2024-01-31
Fix: Call storeUserAnswer on item change (fixes #493)

### LINKED ISSUE #493: ItemsComponentModel not possible to store active item for restoration later
### Subject of the issue
When calling [setActiveItem](https://github.com/adaptlearning/adapt-contrib-core/blob/69044466207afba515edf28c6f5e0a2eac566cbc/js/models/itemsComponentModel.js#L84-L91), if the component is completed, [the storeUserAnswer function doesn't call](https://github.com/adaptlearning/adapt-contrib-core/blob/69044466207afba515edf28c6f5e0a2eac566cbc/js/models/itemsComponentModel.js#L61-L66) and so it is not easy to store the active item if it has changed.

:: Narrative position restoration.

oliverfoster:
fixes #493 

### Fix
* Call storeUserAnswer on item change

### DIFF
--- a/js/models/itemsComponentModel.js
+++ b/js/models/itemsComponentModel.js
@@ -84,10 +84,10 @@ export default class ItemsComponentModel extends ComponentModel {
   setActiveItem(index) {
     const item = this.getItem(index);
     if (!item) return;
-
     const activeItem = this.getActiveItem();
     if (activeItem) activeItem.toggleActive(false);
     item.toggleActive(true);
+    this.storeUserAnswer();
   }
 
 }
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

lemmyadams (approved):

joe-allen-89 (commented):
@oliverfoster if we're firing this.storeUserAnswer on setActiveItem should we also we remove it from checkCompletionStatus as it's currently firing twice when an child's _isVisited changes?

oliverfoster:
_isActive and _isVisited seem to be quite closely linked, _isVisited is used for checkCompletionStatus and _isActive is used for selecting the item, with narrative something can't be _isVisited unless it was once _isActive but the code allows _isVisited and _isActive to be set independently of each other, _isVisited is used to save state here, _isActive is used to keep the is active.

I've just made it save when any property changes to keep it easy.

PR #498, opened 2024-02-02
Fix: Add nav drawer icon override support (fixes #417)

### LINKED ISSUE #417: replace hamburger icon in navbar
The hamburger icon in the navbar opens the Drawer. Ought we consider replacing this icon with a sidebar icon?
sidebar-icon

Rationale:
- The hamburger icon visually represents a vertical list and, in my experience, is overwhelmingly associated with menus.  Consider that the icon used to indicate a list frequently includes "bullets" to distinguish its meaning from the hamburger menu icon. While the Drawer frequently contains lists like Resources and Glossary, the only time it contains a menu is when something like Kineo's TOC is deployed.
- Icons have evolved, and the sidebar icon is finding more use, especially when changing layouts. 
- "Drawer" is a technical label that doesn't get much use anymore. It's more likely to see the term "sidebar" due to its print analog or "sheet". All these terms broadly denote a containing structure. A menu icon feels too narrow for a structure that rarely ever contains an actual menu. If there were an icon that was recognized to mean "resources" or "supplements" or "accessories", that might be a bit more enticing than a sidebar icon. But a menu icon actually feels misleading.

@StuartNicholls @guywillis @kirsty-hames

guywillis:
Fixes: #417 

### Fix
* Add support for overriding nav drawer btn icon in config.json
* Convert nav btn label margin to only apply to labels that are a sibling of an icon
* Added 4 new sidebar icons

Preview of sidebar icons:
Screenshot 2024-02-07 at 16 52 29

Dependency: https://github.com/adaptlearning/adapt_framework/pull/3513


cahirodoherty-learningpool (approved):
👀

templates/nav.hbs
```diff
@@ -20,8 +21,12 @@
   {{/each}}
 
   <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{_config._drawer._position}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
```
oliverfoster, line 23: Does this not become `btn-text` if there is no icon?

schema/config.model.schema
```diff
@@ -349,6 +349,15 @@
               "right"
             ]
           }
+        },
+        "_iconClass": {
+          "type": "string",
+          "required": false,
+          "default": "icon-list",
+          "title": "Drawer icon class",
+          "inputType": "Text",
```
joe-allen-89, line 358: Are we ever likely to use custom _iconClasses? If not is it worth changing this to a select with all available options, a bit easier/nicer to use in authoring tool.

oliverfoster: We wouldn't be able to customise the lists if we change the icon set for a custom theme.
We'd have to have a list containing all icons or all relevant icons from in the vanilla font
We'd have to maintain separate lists for all places where '_iconClass' is going to be added
We'd have to main those list if any other additional icons are added.
Schema enums cannot yet be reused.

joe-allen-89 (approved):
👍 Working nicely

lemmyadams (approved):

eleanor-heath (approved):

templates/nav.hbs
guywillis: Technically, it should yes. But I have taken the approach of leaving as is for the time being to reduce potential issues. 

If we were to change the button class to `btn-text` in the navigation and have the theme set up to have a different font for buttons than body, then we could run into an issue whereby the fonts are different between buttons in the same space.

Screenshot 2024-02-19 at 12 29 34

PR #506, opened 2024-03-15
Fix: Support level course in a11y.ariaLevel (fixes #453)

### LINKED ISSUE #453: `calculateLevel` issue when returning to a completed course
If a completed course is relaunched, the following error occurs when the menu is rendered, and will prevent page loading from the menu:
```JavaScript
adapt.js:141 Uncaught TypeError: Cannot read properties of undefined (reading 'replace')
    at AdaptSingleton.parseRelativeString (adapt.js:141:8)
    at calculateLevel (a11y.js:217:40)
    at calculateLevel (a11y.js:236:14)
    at A11y.ariaLevel (a11y.js:238:12)
    at Object.a11y_aria_level (helpers.js:345:32)
    at Object.c (handlebars.min.js:27:19853)
    at Object.main (templates.js:389:120)
    at c (handlebars.min.js:27:15429)
    at d (handlebars.min.js:27:15744)
    at Object.<anonymous> (templates.js:26:17)
```
Only seems to occur for courses with a role selector, but is due to the completion checking which such plugins perform after modifying the availibility of models when the course loads. When the completion changes, [`updateAria()`](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/views/headingView.js#L23C1-L29) gets called but errors due to no `_ariaLevel` being defined for the 'course' `_type` - https://github.com/adaptlearning/adapt_framework/blob/master/src/course/config.json#L14-L24.

Why do we convert 'course' to 'menu' in https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/views/headingView.js#L18 rather than just adding `"_course": 1` in _config.json_?

May be related to https://github.com/adaptlearning/adapt-contrib-core/issues/355.

oliverfoster:
fixes #453 

### Fix
* Moved course to menu conversion back in the execution hierarchy to allow [a11y.ariaLevel](https://github.com/adaptlearning/adapt-contrib-core/blob/8b36e3ee68941b62e711198f49fa9345c1758132/js/a11y.js#L187) to convert course type to menu instead of relying on [headingView](https://github.com/adaptlearning/adapt-contrib-core/blob/8b36e3ee68941b62e711198f49fa9345c1758132/js/views/headingView.js#L18) to do it.

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -189,6 +189,7 @@ class A11y extends Backbone.Controller {
     level = '1',
     override = null
   } = {}) {
+    if (level === 'course') level = 'menu';
     if (arguments.length === 2) {
       // backward compatibility
       level = arguments[0];
@@ -443,7 +444,7 @@ class A11y extends Backbone.Controller {
 
     const isInDOM = Boolean($element.parents('body').length);
     if (!isInDOM) return false;
-    
+
     const isNotVisible = $branch.toArray().some(item => {
       const style = window.getComputedStyle(item);
       // make sure item is not explicitly invisible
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #508, opened 2024-03-15
Fix: Missing object property lookup restoreUserAnswers, getFeedback, ItemsQuestionModel (fixes #507)

### LINKED ISSUE #507: Defensive code for poor state restoration
### Subject of the issue
There are some instances where the code assumes object properties have values but where the code should return early when the values are missing and the code does currently not return early.

oliverfoster:
fixes #507 

### Fix
* missing _userAnswer in restoreUserAnswers
* getFeedback missing activeItem


swashbuck (approved):
👍

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

Issue #483, opened 2024-01-09
Notify role="dialog" requires a label


kirsty-hames:
### Subject of the issue
Notify uses `dialog` role to help assistive technologies identify the dialog's content as being grouped and separated from the rest of the page content. Adding `role="dialog"` alone is not sufficient to make a dialog accessible and must be properly labeled. See [MDN ARIA: dialog role description for reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#description).

Both [NotifyPopupView](https://github.com/adaptlearning/adapt-contrib-core/blob/6657b362f59cccbb71e49b0e306227fd7c25ddce/js/views/notifyPopupView.js#L17) and [NotifyPushView](https://github.com/adaptlearning/adapt-contrib-core/blob/6657b362f59cccbb71e49b0e306227fd7c25ddce/js/views/notifyPushView.js#L17) use `aria-labelledby` and reference the Notify title to label the dialog however title properties aren't set as required and are sometimes removed in favour of having body text only.

In the instance no title is set, the `dialog` role isn't identified and the user is unaware the content is grouped/separate from the main page content.

Notify is utilised by core and plugins. See below (and please add any I've missed):

- Core question model
- Bookmarking
- LanguagePicker
- Hotgraphic
- Narrative
- Hotgrid
- Visua11y

#### Example using MCQ feedback

When `_feedback title` is set, `.notify__title` is used to label the popup `dialog`. Which reads _"Feedback dialog"_.

<img width="1133" alt="notify_with_title" src="https://github.com/adaptlearning/adapt-contrib-core/assets/7045330/704116ee-1cb7-4f7d-a3c4-6addbcbf46f2">

When no `_feedback title` is set,`.notify__title` is still referenced by `aria-labelledby` but the element doesn't exist so the popup `dialog` has no label. _"dialog"_ isn't announced.

<img width="1139" alt="notify_without_title" src="https://github.com/adaptlearning/adapt-contrib-core/assets/7045330/03da56b3-329b-46c9-ac77-aaa6e23a38e8">

### Expected behaviour
I think most content authors won't be aware of the accessibility implications by not setting a title so either highlighting this in the instruction/help text, setting titles as required or providing a fallback should prevent this.

kirsty-hames:
**Core question model**
[...]

PR #428, opened 2023-08-11
Fix: remove component diplayTitle and Title fallbacks for feedbackTitle

joe-allen-89:
[//]: # (Please title your PR according to eslint commit conventions)
[//]: # (See https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint#eslint-convention for details)

[//]: # (Link the PR to the original issue)
Currently if no feedback title is given the questionModel will return the components display title/title meaning that the altTitle isn't used unless all of the feedback/component titles are empty. 

My suggestion is to remove the components display title/title as fallbacks so if no feedback title is given the altTitle is used for screen readers with no visible feedback title. 

Original issue - #411 

[//]: # (Delete Fix, Update, New and/or Breaking sections as appropriate)
### Fix
* remove component diplayTitle and Title fallbacks for feedbackTitle (fixes #411 )

PR #416, opened 2023-07-26
Fix: PartlyCorrect feedback fallback values

cahirodoherty-learningpool:
[//]: # (Please title your PR according to eslint commit conventions)
[//]: # (See https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint#eslint-convention for details)

[//]: # (Link the PR to the original issue)
Fixes #415 

[//]: # (Delete Fix, Update, New and/or Breaking sections as appropriate)
### Fix
* 'Partly Correct Not Final' feedback should fall back to 'Partly Correct Final' which in turn should fall back to 'Incorrect Final' as per instruction text

PR #201, opened 2022-07-27
Update: provide alternative heading for AT (fixes #173)

chris-steele:
[//]: # (Link the PR to the original issue)
fixes https://github.com/adaptlearning/adapt-contrib-core/issues/173

### Update
* Provide an alternative heading for Assistive Technology if no visual heading is required

PR #510, opened 2024-03-19
Fix: Fallback value of title used for component feedback when _feedback.title not set (509)

### LINKED ISSUE #509: Component feedback title not falling back to component title
Describe your issue here.

### Your environment
* FW v5.36.2 (Core v6.45.6)

### Steps to reproduce
Use MCQ on a course and leave the Feedback title blank.

### Expected behaviour
The help text indicates that component title will be used as if this field is left empty

### Actual behaviour
No title displayed on component feedback

cahirodoherty-learningpool:

Fixes #509 

### Fix
* Fallback value of title used for component feedback when _feedback.title not set

### DIFF
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -288,7 +288,7 @@ class QuestionModel extends ComponentModel {
         Adapt.course.get('_globals')._accessibility.altFeedbackTitle ||
         '',
       title: feedback.title ||
-        this.get('displayTitle') ||
+        this.get('title') ||
         '',
       _classes: feedback._classes,
       ...(isLegacyConfig
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -287,8 +287,9 @@ class QuestionModel extends ComponentModel {
     const title = feedback.title || this.get('title') || altTitle || '';
 
     const feedbackConfig = {
+      isAltTitle: (title === altTitle),
       altTitle,
-      title,
+      title: Handlebars.compile(title)(this.toJSON()),
       _classes: feedback._classes,
       ...(isLegacyConfig
         ? getLegacyConfigObject()
@@ -296,7 +297,7 @@ class QuestionModel extends ComponentModel {
       )
     };
 
-    if (feedbackConfig?._graphic?._src && !feedbackConfig?._imageAlignment) {
+    if (feedbackConfig._graphic?._src && !feedbackConfig._imageAlignment) {
       feedbackConfig._imageAlignment = 'right';
     }
 
@@ -307,10 +308,8 @@ class QuestionModel extends ComponentModel {
   setupFeedback() {
     if (!this.has('_feedback')) return;
     const { altTitle = '', title = '', body = '' } = this.getFeedback();
-    const isAltTitle = title === altTitle;
 
     this.set({
-      isAltTitle,
       feedbackTitle: Handlebars.compile(title)(this.toJSON()),
       feedbackMessage: Handlebars.compile(body)(this.toJSON())
     });
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -283,13 +283,12 @@ class QuestionModel extends ComponentModel {
       ) || {};
     };
 
+    const altTitle = feedback.altTitle || Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
+    const title = feedback.title || this.get('title') || altTitle || '';
+
     const feedbackConfig = {
-      altTitle: feedback.altTitle ||
-        Adapt.course.get('_globals')._accessibility.altFeedbackTitle ||
-        '',
-      title: feedback.title ||
-        this.get('title') ||
-        '',
+      altTitle,
+      title,
       _classes: feedback._classes,
       ...(isLegacyConfig
         ? getLegacyConfigObject()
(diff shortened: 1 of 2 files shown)


oliverfoster:
Could we resolve these issues please?
@kirsty-hames @joe-allen-89 @cahirodoherty-learningpool 

These fallback keep coming and going:
* Added originally by Cahir https://github.com/adaptlearning/adapt-contrib-core/pull/416
* Removed by Kirsty and Joe https://github.com/adaptlearning/adapt-contrib-core/pull/428

cahirodoherty-learningpool:
> It seems it must be possible to leave the feedback `title` empty in order for `altTitle` to work. Invariably most components will have either a `displayTitle` or a `title` so if either property is used as a default for feedback title, then `altTitle` never gets applied.
Yes, I'm a little confused as to the actual code around `altTitle` and `altFeedbackTitle`. `altFeedbackTitle` is the schema property, but I don't think the code actually ever reads that; Instead we have https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/models/questionModel.js#L287, but `altTitle` won't exist to be read at that point. I think its just a typo, and not central to your argument here Oliver, but worth noting.

> Why should the feedback title default to the question `displayTitle` or `title`?
This is driven by user expectations from how it was working before and the helper text found on many components, e.g. https://github.com/adaptlearning/adapt-contrib-mcq/blob/master/properties.schema#L243

kirsty-hames:
> Yes, I'm a little confused as to the actual code around `altTitle` and `altFeedbackTitle`. `altFeedbackTitle` is the schema property, but I don't think the code actually ever reads that; Instead we have https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/models/questionModel.js#L287, but `altTitle` won't exist to be read at that point. I think its just a typo, and not central to your argument here Oliver, but worth noting.

`altFeedbackTitle` is a global title, set once in _course.json_ rather than per component feedback. `altTitle` is set per question feedback and overrides `altFeedbackTitle`. Neither title display on screen and are used by screen readers.

Notify requires a title as this is used to [label](https://github.com/adaptlearning/adapt-contrib-core/blob/6657b362f59cccbb71e49b0e306227fd7c25ddce/js/views/notifyPopupView.js#L17) the Notify 'dialog' popup (see [ticket for ref](https://github.com/adaptlearning/adapt-contrib-core/issues/483)).

If we're providing a fallback title (`title` and/or `displayTitle`), then `altFeedbackTitle` and `altTitle` should be removed as these won't get applied. Before removing, is there a use case for keeping alt titles? e.g. if using inline feedback displaying a title for the question and feedback can become repetitive etc. I couldn't find the issue/PR for why `altTitle` got added in the first place but I have seen these used in projects in favor of displaying a title. Do people have a preference?

oliverfoster:
@kirsty-hames 

`altTitle` original [pr](https://github.com/adaptlearning/adapt-contrib-core/pull/201) and [issue](https://github.com/adaptlearning/adapt-contrib-core/issues/173)

Including @chris-steele

cahirodoherty-learningpool:
> > Yes, I'm a little confused as to the actual code around `altTitle` and `altFeedbackTitle`. `altFeedbackTitle` is the schema property, but I don't think the code actually ever reads that; Instead we have https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/models/questionModel.js#L287, but `altTitle` won't exist to be read at that point. I think its just a typo, and not central to your argument here Oliver, but worth noting.
> 
> `altFeedbackTitle` is a global title, set once in _course.json_ rather than per component feedback. `altTitle` is set per question feedback and overrides `altFeedbackTitle`. Neither title display on screen and are used by screen readers.
> 
> Notify requires a title as this is used to [label](https://github.com/adaptlearning/adapt-contrib-core/blob/6657b362f59cccbb71e49b0e306227fd7c25ddce/js/views/notifyPopupView.js#L17) the Notify 'dialog' popup (see [ticket for ref](https://github.com/adaptlearning/adapt-contrib-core/issues/483)).
> 
> If we're providing a fallback title (`title` and/or `displayTitle`), then `altFeedbackTitle` and `altTitle` should be removed as these won't get applied. Before removing, is there a use case for keeping alt titles? e.g. if using inline feedback displaying a title for the question and feedback can become repetitive etc. I couldn't find the issue/PR for why `altTitle` got added in the first place but I have seen these used in projects in favor of displaying a title. Do people have a preference?

Taking a look at the hbs file, it seems we are reading the models `altTitle` in the elseif. I think this should actually be `altFeedbackTitle` as that is the custom property pushed onto the model based on feedback.altTitle || _globals.altFeedbackTitle if we are going to keep altTitle. Also, the difference in the title and altTitle html is minimal. Perhaps we can move all the setting of display logic into the View to help clarify our fallbacks.

I'll add this to the current PR for a bit of a cleanup, but we should definitely resolve what our fallbacks are and in which order

cahirodoherty-learningpool:
I've updated the PR to simplify the hbs by pushing display logic more on the view. We still have the decision to make over the fields we sequentially fall back on for the feedbackTitle field that gets resolved here though, but hopefully it's now easier to see that order

kirsty-hames:
> I've updated the PR to simplify the hbs by pushing display logic more on the view. We still have the decision to make over the fields we sequentially fall back on for the feedbackTitle field that gets resolved here though, but hopefully it's now easier to see that order

Thanks @cahirodoherty-learningpool, this is much clearer. When I update to your latest [commit]([4bd9b57](https://github.com/adaptlearning/adapt-contrib-core/pull/510/commits/4bd9b57be3f29c7665ea59910e80c0fb585200f4)) however no title is rendering for me regardless which title I set (title, altTitle or altFeedbackTitle).

In terms of order, I'd propose the following:
`_feedback.title` with default of 'Feedback'. If left empty, no visual title will display, only an `aria-label` title will render (altTitle or altFeedbackTitle).
`_feedback.altTitle` with no default. Optional if question specific alt text is required.
`_globals._accessibility.altFeedbackTitle` with default of 'Feedback'. Ensures the Feedback always has a label for accessibility.

Reviewing some recent courses, I find feedback title text is one of the following:
Feedback,
Content specific title,
Blank (with an aria-label 'Feedback').

If we want to use the component `title`/`displayTitle` as a fallback I'm not sure how we would add this other than providing a config option. If so, is using `title`/`displayTitle` as a fallback typically done globally for a course or per question instance?

joe-allen-89 (approved):
👍

PR #521, opened 2024-04-23
Mandatory and optional labels added to schema (fixes #520)

### LINKED ISSUE #520: Global ARIA labels for required and optional content
Translatable labels for required or optional pages/articles/blocks/components.

### Expected behaviour
So we can better display this status. This was intended for menu items and page headers, but it should be usable in all content types.

zubairslamdien:
Fixes #520 

### New
* New "mandatory" and "optional" labels added to _ariaLabels in the course schema. Updates to the page template and Boxmenu will follow.

### DIFF
--- a/schema/course.model.schema
+++ b/schema/course.model.schema
@@ -339,10 +339,10 @@
                   "required": true,
                   "translatable": true
                 },
-                "mandatory": {
+                "required": {
                   "type": "string",
                   "title": "",
-                  "default": "Mandatory",
+                  "default": "Required",
                   "inputType": "Text",
                   "required": true,
                   "translatable": true
--- a/schema/course.model.schema
+++ b/schema/course.model.schema
@@ -338,6 +338,22 @@
                   "inputType": "Text",
                   "required": true,
                   "translatable": true
+                },
+                "mandatory": {
+                  "type": "string",
+                  "title": "",
+                  "default": "Mandatory",
+                  "inputType": "Text",
+                  "required": true,
+                  "translatable": true
+                },
+                "optional": {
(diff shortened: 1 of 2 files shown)


schema/course.model.schema
```diff
@@ -338,6 +338,22 @@
                   "inputType": "Text",
                   "required": true,
                   "translatable": true
+                },
+                "mandatory": {
```
oliverfoster, line 342: mandatory vs required?

StuartNicholls (approved):

schema/course.model.schema
StuartNicholls: @oliverfoster I think mandatory is the most commonly used. I guess it's a stronger word? I'll go with consensus though

oliverfoster: Stronger in what sense? More characters and syllables? `required` has a higher scrabble score at 18 vs `mandatory`'s 15.

StuartNicholls: If words had muscles, Mandatory would have bigger ones

oliverfoster: If we go for `"compulsory": "Compulsory"` we could compel people to do the learning rather than just mandate it?

kirsty-hames: I would have thought "required" to be the common use case but I have seen both use cases. The `default` can easily be amended per course when needed.

kirsty-hames (approved):

oliverfoster (approved):

kirsty-hames (approved):

cahirodoherty-learningpool (approved):
👀

PR #524, opened 2024-04-26
Fix: enforcing type boolean on _isActive (Fixes #523)

### LINKED ISSUE #523: itemModel toggleActive function currently not forcing boolean setting
_isActive property always expected to be boolean but toggleActive not strictly enforcing this currently

ethan-lp:
_isActive property always expected to be boolean but toggleActive not strictly enforcing this currently

### Fix
enforcing type boolean on _isActive within toggleActive function

### DIFF
--- a/js/models/itemModel.js
+++ b/js/models/itemModel.js
@@ -15,11 +15,11 @@ export default class ItemModel extends LockingModel {
   }
 
   toggleActive(isActive = !this.get('_isActive')) {
-    this.set('_isActive', isActive);
+    this.set('_isActive', Boolean(isActive));
   }
 
   toggleVisited(isVisited = !this.get('_isVisited')) {
-    this.set('_isVisited', isVisited);
+    this.set('_isVisited', Boolean(isVisited));
   }
 
 }
--- a/js/models/itemModel.js
+++ b/js/models/itemModel.js
@@ -15,7 +15,7 @@ export default class ItemModel extends LockingModel {
   }
 
   toggleActive(isActive = !this.get('_isActive')) {
-    this.set('_isActive', Boolean(isActive));
+    this.set('_isActive', isActive);
   }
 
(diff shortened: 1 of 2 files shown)


js/models/itemModel.js
```diff
@@ -15,7 +15,7 @@ export default class ItemModel extends LockingModel {
   }
 
   toggleActive(isActive = !this.get('_isActive')) {
-    this.set('_isActive', isActive);
+    this.set('_isActive', !!isActive);
```
oliverfoster, line 18:
```suggestion
    this.set('_isActive', Boolean(isActive));
```

oliverfoster (changes requested):

js/models/itemModel.js
oliverfoster: Why is this needed? Which bit of code is trying to set it to a non-boolean value?

ethan-lp: `restoreUserAnswers()` within the `itemsQuestionModel.js` file, we are seeing it is passing a binary value and not boolean.

oliverfoster: https://github.com/adaptlearning/adapt-contrib-core/blob/de6fcd9c58ad7ed0d793920b048fcb213bdcc980/js/models/itemsQuestionModel.js#L47

Could you please wrap that line in a Boolean() call? As it's the offending cause.

ethan-lp: Yeah done and understood. Thanks Oliver

cahirodoherty-learningpool (approved):
👀

js/models/itemModel.js
oliverfoster: I'm happy for you to leave your original bit in also in toggleActive.

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

lemmyadams (approved):

PR #532, opened 2024-05-03
Fix: exclude non-trackable models from `checkCompletionStatusFor`

### LINKED ISSUE #526: `checkCompletionStatusFor` doesn't complete if model and its children are optional and trickled
If a model and its children are `"_isOptional": true` and set to use a trickle button, the model does not mark as `_isComplete` because the trickle button is not visually available to be completed.

Whilst the JSON is unlikely to be configured this way whilst authoring, it is possible when using a role selector or diagnostic type plugin which can dynamically update `_isOptional` depending on user choices.

This can be resolved by excluding non-trackable components such as the trickle button from: https://github.com/adaptlearning/adapt-contrib-core/blob/de6fcd9c58ad7ed0d793920b048fcb213bdcc980/js/models/adaptModel.js#L316

It might be useful to create an API method returning these models so it can be used elsewhere as required:
```JavaScript
return this.getAvailableChildModels().filter(child => child.get('_isTrackable') !== false);
```

danielghost:
Fixes #526.

### Fix
* Exclude non-trackable models from `checkCompletionStatusFor`.

### New
* Added `getAvailableTrackableChildModels` to return only `_isAvailable: true` and `_isTrackable !== false` models.

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -313,7 +313,7 @@ export default class AdaptModel extends LockingModel {
    */
   checkCompletionStatusFor(completionAttribute = '_isComplete') {
     let completed = false;
-    const children = this.getAvailableChildModels();
+    const children = this.getAvailableTrackableChildModels();
     const requireCompletionOf = this.get('_requireCompletionOf');
     const isOptional = this.get('_isOptional');
     const isEveryChildOptional = children.every(child => child.get('_isOptional'));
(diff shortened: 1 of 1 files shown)


joe-allen-89 (approved):
👍

oliverfoster:
_isTrackable should not be intertwined with completion or optionality, please fix this in trickle rather than core.

ethan-lp (approved):

lemmyadams (approved):

Issue #512, opened 2024-03-27
Migrate drawer button to Navigation Button API


swashbuck:
### Subject of the enhancement
Migrate the drawer button to [Navigation Button API](https://github.com/adaptlearning/adapt-contrib-core/pull/339). This will allow us to modify the button label text separately from the aria label.

Currently, it is hard-coded into the [nav.hbs](https://github.com/adaptlearning/adapt-contrib-core/blob/master/templates/nav.hbs#L23) template.

PR #536, opened 2024-05-13
Fix: Add default values for Drawer button icon and position (fixes #534)

### LINKED ISSUE #534: Add default values for Drawer button icon and position
### Subject of the issue
In #498, the option to override the Drawer icon was introduced. For older courses upgrading to a newer version of `adapt-contrib-core`, the icon will be missing since there is no default value.

Unless there is a need to have the option of a nav button without an icon (e.g. when using nav text labels only), I think we should add a default value of `icon-list` with a position of `auto`.

### Your environment
* Core v6.46.7
* FW v5.38.2

### Screenshots
drawer-icon-missing

swashbuck:
Fixes #534 

**Note:** This PR is considered a stopgap until #512 is completed.

### Fix
Sets defaults for the Drawer navigation button:
* Set a default [icon](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Icons#icons-in-adapt) of `icon-list` for the `_iconClass` property. The icon will no longer be hidden if `_iconClass` is not set.
* Set a default `_position` of `auto`

### Testing
In _config.json_, configure the `_drawer` object with _and_ without values for `_iconClass` and `_position`. When not set, it should show the list icon and the position should be `auto`.

list-icon

```
  "_drawer": {
    "_showEasing": "easeOutQuart",
    "_hideEasing": "easeInQuart",
    "_duration": 400,
    "_position": "right",
    "_iconClass": "icon-folder"
  },
```

### DIFF
--- a/templates/nav.hbs
+++ b/templates/nav.hbs
@@ -21,7 +21,7 @@
   {{/each}}
 
   <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{#if _config._drawer._position}}{{_config._drawer._position}}{{/if}}{{#unless _config._drawer._position}}auto{{/unless}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
-    <span class="icon {{#if _config._drawer._iconClass}}{{_config._drawer._iconClass}}{{/if}}{{#unless _config._drawer._iconClass}}icon-list{{/unless}}" aria-hidden="true"></span>
+    <span class="icon {{#if _config._drawer._iconClass}}{{_config._drawer._iconClass}}{{else}}icon-list{{/if}}" aria-hidden="true"></span>
     {{#if Adapt.course._navigation._showLabel}}
     <span class='nav__btn-label' aria-hidden="true">{{_globals._accessibility._ariaLabels.navigationDrawer}}</span>
     {{/if}}
--- a/templates/nav.hbs
+++ b/templates/nav.hbs
@@ -20,7 +20,7 @@
     <div class="nav__spacer" aria-hidden="true" data-order="{{_navOrder}}"></div>
   {{/each}}
 
-  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{#if _config._drawer._position}}{{_config._drawer._position}}{{/if}}{{#unless _config._drawer._position}}auto{{/unless}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
+  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{#if _config._drawer._position}}{{_config._drawer._position}}{{else}}auto{{/if}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

templates/nav.hbs
```diff
@@ -20,10 +20,8 @@
     <div class="nav__spacer" aria-hidden="true" data-order="{{_navOrder}}"></div>
   {{/each}}
 
-  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{_config._drawer._position}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
-    {{#if _config._drawer._iconClass}}
-    <span class="icon {{_config._drawer._iconClass}}" aria-hidden="true"></span>
-    {{/if}}
+  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{#if _config._drawer._position}}{{_config._drawer._position}}{{/if}}{{#unless _config._drawer._position}}auto{{/unless}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
+    <span class="icon {{#if _config._drawer._iconClass}}{{_config._drawer._iconClass}}{{/if}}{{#unless _config._drawer._iconClass}}icon-list{{/unless}}" aria-hidden="true"></span>
```
oliverfoster, line 24:
```suggestion
    <span class="icon {{#if _config._drawer._iconClass}}{{_config._drawer._iconClass}}{{else}}icon-list{{/if}}" aria-hidden="true"></span>
```

https://handlebarsjs.com/guide/block-helpers.html#conditionals

You happy this this in the handlebars rather than the javascript?
https://github.com/adaptlearning/adapt-contrib-core/blob/7b126ae7bcce7ea22d2ecdf57069b4b0b2826feb/js/views/navigationView.js#L80

oliverfoster (changes requested):

templates/nav.hbs
swashbuck: > You happy this this in the handlebars rather than the javascript?

@oliverfoster I think so, yeah. It's just a stopgap solution, and it will be easy to remove once we use Navigation Button API.

templates/nav.hbs
```diff
@@ -20,10 +20,8 @@
     <div class="nav__spacer" aria-hidden="true" data-order="{{_navOrder}}"></div>
   {{/each}}
 
-  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{_config._drawer._position}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
-    {{#if _config._drawer._iconClass}}
-    <span class="icon {{_config._drawer._iconClass}}" aria-hidden="true"></span>
-    {{/if}}
+  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{#if _config._drawer._position}}{{_config._drawer._position}}{{/if}}{{#unless _config._drawer._position}}auto{{/unless}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
```
oliverfoster, line 23:
```suggestion
  <button name="drawer" class="btn-icon nav__btn nav__drawer-btn js-nav-drawer-btn is-position-{{#if _config._drawer._position}}{{_config._drawer._position}}{{else}}auto{{/if}}" data-event="toggleDrawer" aria-expanded="false" aria-label="{{_globals._accessibility._ariaLabels.navigationDrawer}}" data-order="{{_globals._extensions._drawer._navOrder}}" data-tooltip-id="drawer">
```

oliverfoster (changes requested):

oliverfoster (approved):

guywillis (approved):
👀

PR #539, opened 2024-05-15
New: Allow any component to use mobileBody for smaller devices (fixes #538)

### LINKED ISSUE #538: Mobile body and instruction behave differently
### Subject of the issue
When `_isNarrativeOnMobile` is `true`, both `mobileBody` and `mobileInstruction` are correctly shown on mobile. However, when `_isNarrativeOnMobile` is `false`, only `mobileInstruction` is shown on mobile and the body text uses `body` (not `mobileBody`).

`mobileBody` should always be shown on mobile view regardless of the value of `_isNarrativeOnMobile`.

We should also check whether the issue is present or caused by the Narrative plugin.

swashbuck:
Fix #538 

### New
* Introduces the `mobileBody` property for component headers

### Testing
1. Add a `body` and `mobileBody` value to any component.
2. Switch to mobile view (less than medium breakpoint) and ensure that body text uses `mobileBody` instead of `body`.
3. Switch back to desktop view (medium breakpoint or higher) and ensure that body text uses `body` instead of `mobileBody`.

### Future consideration
Add `mobileBody` and `mobileInstruction` to the core component schemas. Then, remove them from the schemas of components like [Hot Graphic](https://github.com/adaptlearning/adapt-contrib-hotgraphic/blob/55b26b7fdae55f6c3d97384da2bd39e2d45e6616/schema/component.schema.json#L78) and [Narrative](https://github.com/adaptlearning/adapt-contrib-narrative/blob/c350f7610a0ee927a1259efdd2623e46e2847e94/schema/component.schema.json#L35). Would require migration scripts.

### DIFF
--- a/templates/header.jsx
+++ b/templates/header.jsx
@@ -5,11 +5,12 @@ import { classes, prefixClasses, compile } from 'core/js/reactHelpers';
 
 /**
  * Content header for displayTitle, body, instruction text, etc.
- * instruction and mobileInstruction will switch automatically
+ * instruction / mobileInstruction and body / mobileBody will switch automatically
  * @param {Object} props
  * @param {string} [props.displayTitle]
  * @param {string} [props.body]
  * @param {string} [props.instruction]
+ * @param {string} [props.mobileBody]
  * @param {string} [props.mobileInstruction]
  * @param {string} [props._type]
  * @param {string} [props._component]
--- a/templates/header.jsx
+++ b/templates/header.jsx
@@ -5,7 +5,7 @@ import { classes, prefixClasses, compile } from 'core/js/reactHelpers';
 
 /**
  * Content header for displayTitle, body, instruction text, etc.
- * instruction / mobileInstruction and body / mobileBody will switch automatically
+ * body / mobileBody and instruction / mobileInstruction will switch automatically
  * @param {Object} props
  * @param {string} [props.displayTitle]
  * @param {string} [props.body]
--- a/templates/header.jsx
+++ b/templates/header.jsx
@@ -22,6 +22,7 @@ export default function Header(props) {
     displayTitle,
     body,
     instruction,
+    mobileBody,
     mobileInstruction,
     _type,
     _component,
@@ -33,6 +34,9 @@ export default function Header(props) {
       _extension && _extension.toLowerCase()
     ].filter(Boolean)
   } = props;
+  const sizedBody = (mobileBody && !device.isScreenSizeMin('medium')) ?
+    mobileBody :
+    body;
   const sizedInstruction = (mobileInstruction && !device.isScreenSizeMin('medium')) ?
     mobileInstruction :
     instruction;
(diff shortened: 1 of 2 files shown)


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

kirsty-hames (commented):
This works as expected thanks @swashbuck. 

Should we do the same for _component.hbs_ [`body`](https://github.com/adaptlearning/adapt-contrib-core/blob/d4a8f277f05a1eafc4c6a20fe188a7e3c0a9ed88/templates/partials/component.hbs#L16)? I think Core wise, Media is the only component still using _hbs_ templates.

kirsty-hames:
> Add `mobileBody` and `mobileInstruction` to the core component schemas. Then, remove them from the schemas of components like [Hot Graphic](https://github.com/adaptlearning/adapt-contrib-hotgraphic/blob/55b26b7fdae55f6c3d97384da2bd39e2d45e6616/schema/component.schema.json#L78) and [Narrative](https://github.com/adaptlearning/adapt-contrib-narrative/blob/c350f7610a0ee927a1259efdd2623e46e2847e94/schema/component.schema.json#L35). Would require migration scripts.

Does this need raising as a separate issue so it's not forgotten when the PR is merged?

swashbuck:
> Does this need raising as a separate issue so it's not forgotten when the PR is merged?

@kirsty-hames Great idea. I've created issues for these two plugins.

https://github.com/adaptlearning/adapt-contrib-hotgraphic/issues/308
https://github.com/adaptlearning/adapt-contrib-narrative/issues/302

swashbuck:
> Should we do the same for component.hbs [body](https://github.com/adaptlearning/adapt-contrib-core/blob/d4a8f277f05a1eafc4c6a20fe188a7e3c0a9ed88/templates/partials/component.hbs#L16)? I think Core wise, Media is the only component still using hbs templates.

@kirsty-hames I've added `mobileBody` to _component.hbs_. However, it appears that the expectation is for individual plugins to determine whether `body` or `mobileBody` would be used ([same](https://github.com/adaptlearning/adapt-contrib-core/blob/4d6f06e741c0de2768cbc66ff185dd3bb0ebc292/templates/partials/component.hbs#L28) for `instruction` and `mobileInstruction`). This is [different](https://github.com/adaptlearning/adapt-contrib-core/blob/4d6f06e741c0de2768cbc66ff185dd3bb0ebc292/templates/header.jsx#L36) from the _header.jsx_ template which determines the appropriate property to use within the template itself.

So, for this to work in Media and other plugins that still use the older template, updates to the Media plugin's JS would be needed. The change in [2b565d8](https://github.com/adaptlearning/adapt-contrib-core/pull/539/commits/2b565d814eec853ad18c4af6ef68a7ce6ebcd0c9) will help to facilitate this, though.

kirsty-hames (approved):

PR #541, opened 2024-05-17
Review

oliverfoster:
### transitionsEnded
When `!willTransition`,the return early can be moved to the top of the function before the promise and nested function instantiation.
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L24

### getActiveTransitions & getTransitionsFromCSS
The word Transitions in these names refers to two different concepts. The first is an array of `CSSTransition` objects and the second is an array of `{ transitionProperty, timeToEnd }` objects.

### getActiveTransitions & willTransition
`getActiveTransitions` is only used inside `willTransition`, there is probably no need to split this into two functions as the code never repeats.
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L85-L96
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L36-L45

### willTransition, getResolutionDuration & getTransitionsFromCSS
These functions perform the same behaviour for most of the code. Suggest consolidating the common functionality with  `getTransitionsFromCSS` into a `getTransitionsLongestEndTime` function instead.
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L68-L83
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L98-L112
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L47-L66
The bottom of `getResolutionDuration` is more for offsetting the timeout to run at a goodly position after transition start, it should probably go next to the `setTimeout` call.
https://github.com/adaptlearning/adapt-contrib-core/blob/1e5eb32ee87307768c490936c8a4d71b34614e94/js/transitions.js#L110-L111

### jsDoc comments
* I've changed `String` to `string` as `String` is an instance of `new String()` and `string` is a literal
* I've changed `property` to `[property=null]` as that shows and optional property with a default value

### DIFF
--- a/js/transitions.js
+++ b/js/transitions.js
@@ -1,15 +1,71 @@
 import logging from 'core/js/logging';
 
 /**
+ * Returns whether `CSSTransitions` are still active for an element.
+ * An optional `transition-property` can be specified, else all properties will be evaluated.
+ * @param {jQuery} $element
+ * @param {string} [property=null]
+ * @returns {Boolean}
+ */
+function hasActiveTransition($element, property = null) {
+  let element = $element;
+  if (element instanceof $) element = element[0];
+  let transitions = element.getAnimations().filter(animation => animation instanceof window.CSSTransition);
+  if (property) transitions = transitions.filter(({ transitionProperty }) => transitionProperty === property);
+  return Boolean(transitions.length);
+}
+
+/**
+ * Returns the longest end time of the configured transitions
+ * @param {jQuery} $element
+ * @param {string} [property=null]
+ * @returns {Boolean}
+ */
+function getTransitionsLongestEndTime($element, property = null) {
+  const properties = $element.css('transition-property').split(',').map(property => property.trim());
+  const durations = $element.css('transition-duration').split(',').map(duration => parseFloat(duration));
+  const delays = $element.css('transition-delay').split(',').map(delay => parseFloat(delay));
+  const endTimes = properties.reduce((result, transitionProperty, index) => {
+    // cycle through populated timings as per spec
+    const duration = durations[index % durations.length];
+    const delay = delays[index % delays.length];
+    const endTime = duration + delay;
+    result[transitionProperty] = endTime;
+    return result;
+  }, {});
+  const isAllProperty = property === 'all' || Boolean(endTimes.all);
+  const longestEndTime = (property && !isAllProperty)
+    ? endTimes[property]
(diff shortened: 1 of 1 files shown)


js/transitions.js
```diff
@@ -21,101 +25,70 @@ export async function transitionsEnded($element, property = null) {
...
-    }
-  });
+    const endTime = duration + delay;
+    result[transitionProperty] = endTime;
+    return result;
+  }, {});
+  const isAllProperty = property === 'all' || Boolean(endTimes.all);
+  const longestEndTime = (property && !isAllProperty)
+    ? endTimes[property]
+    : Object.values(endTimes).sort((a, b) => a.timeToEnd - b.timeToEnd).pop();
```
danielghost, line 58: `timeToEnd` no longer exists, so this will just return the last item in the Array without sorting. As the object now uses the property name as the key, this could be changed to:

```suggestion
    : Math.max(...Object.values(endTimes));
```

danielghost (changes requested):

js/transitions.js
danielghost: > It's possible to log the element instead of having to derive an id or name.

I thought printing the whole element would be too verbose, but as this is now in a second argument I guess it adds more detail and can be expanded to reveal the element as required.

> getActiveTransitions is only used inside willTransition, there is probably no need to split this into two functions as the code never repeats.

This makes sense for the code within `transitionsEnded`, but I had purposefully split these up should they need be useful by other plugins in future. Fine for these to be merged though as probably unlikley to be needed.

oliverfoster: `getActiveTransitions` is pretty much just the native `getAnimations` with the `CSSTransition` class check, so not much hassle to add late or recreate.

oliverfoster: > `timeToEnd` no longer exists, so this will just return the last item in the Array without sorting. As the object now uses the property name as the key, this could be changed to:

Ominous bit of code to read out loud.

danielghost (approved):

PR #550, opened 2024-06-05
New: Allow for instance template overrides (fixes #549)

### LINKED ISSUE #549: Allow jsx instance template overrides
### Subject of the issue
Current templates must be defined on the class rather than the instance which makes swapping between templates per instance difficult.

### Your environment
* Master FW & AT

### Expected behaviour
To be able to define templates on instances.

guywillis:
Fixes: #549

### New
* Allow for instance template overrides

```js
class TestView extends ComponentView {
  get template() {
    return 'templateName.jsx';
  }
}
```

instead of:

```js
class TestView extends ComponentView {
  // ...
}

TestView.template = 'templateName.jsx';
```

### DIFF
--- a/js/views/adaptView.js
+++ b/js/views/adaptView.js
@@ -24,7 +24,7 @@ class AdaptView extends Backbone.View {
       'change:_isHidden': this.toggleHidden,
       'change:_isComplete': this.onIsCompleteChange
     });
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):
👀 seen working

joe-allen-89 (approved):
👍

PR #551, opened 2024-06-05
Fix: Resolve title/altTitle discrepancy (fixes #535)

### LINKED ISSUE #535: Missing titles for feedback Notify popups
### Subject of the issue
I believe this is a bug, but let me know if it is intended behavior.

If `_feedback.altTitle` is not present, is an empty value, or is the same value as `_feedback.title`, the title is not shown in the popup. However, if `_feedback.altTitle` is different from `_feedback.title` (either different text or just a single space character), then the title appears as expected.

You can test this with the core course from running `adapt create course` and viewing the Question Components topic.

This may be caused by #509 .

### Your environment
* FW 5.38.2
* Core 6.46.7

### Steps to reproduce
Configure the `_feedback` object in one of the following ways:

1. `_feedback.title` is the same as `_feedback.altTitle`
2. `_feedback.altTitle` is removed completely
3. `_feedback.altTitle` is an empty string ""

For example:

```
"_feedback": {
  "title": "Feedback",
  "altTitle": "Feedback",
  "correct": "Correct feedback text.",
  "_incorrect": {
    "notFinal": "",
    "final": "Incorrect feedback text."
  }
},
```

### Expected behaviour
"Feedback" title should appear.

### Actual behaviour
No title appears

swashbuck:
Fixes #535

### Fix
* Rework how `_feedback.title` and `_feedback.altTitle` work. `_feedback.altTitle`, if set, will take precedence and the feedback title will be visually hidden.
* Removes `altTitle` from feedbackConfig() as it did not seem to be used in the Notify templates.

The new behavior is as follows:
1. If `_feedback.altTitle` is set, this will be used as a title aria label. No visual title is displayed.
2. If `_feedback.title` is set and `_feedback.altTitle` is _not_ set, a title is displayed with this value.
3. If _neither_ `_feedback.title` nor `_feedback.altTitle` are set, a title is displayed with the _component_'s title.
4. If none of the above are set, `_globals._accessibility.altFeedbackTitle` is used as an aria label and no visual title is displayed.
5. Otherwise, no title is displayed or used as an aria label

The precedence basically goes:

1. `_feedback.altTitle`
2. `_feedback.title`
3. Component title
4. `_globals._accessibility.altFeedbackTitle`

### Related
This removes the `_feedback.altTitle` from the sample FW course. Otherwise, by default, the sample course does not show a feedback title due to the presence of the alt titles.
https://github.com/adaptlearning/adapt_framework/pull/3567

### DIFF
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -286,7 +286,7 @@ class QuestionModel extends ComponentModel {
     const altTitle = feedback.altTitle || Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
     const hasTitle = Boolean(feedback.title || this.get('title'));
     const isAltTitle = altTitle && !hasTitle;
-    const title = hasTitle ? (feedback.title || this.get('title')) : altTitle || '';
+    const title = (feedback.title || this.get('title') || altTitle || '');
 
     const feedbackConfig = {
       isAltTitle,
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -284,7 +284,7 @@ class QuestionModel extends ComponentModel {
     };
 
     const altTitle = feedback.altTitle || Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
-    const isAltTitle = altTitle;
+    const isAltTitle = Boolean(altTitle);
     const title = (feedback.title || this.get('title') || altTitle || '');
 
     const feedbackConfig = {
--- a/js/models/questionModel.js
+++ b/js/models/questionModel.js
@@ -283,13 +283,12 @@ class QuestionModel extends ComponentModel {
       ) || {};
     };
 
-    const altTitle = feedback.altTitle || Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
-    const isAltTitle = Boolean(altTitle);
-    const title = (feedback.title || this.get('title') || altTitle || '');
+    const hasTitle = Boolean(feedback.title || this.get('title'));
+    const isAltTitle = Boolean(feedback.altTitle) || (!hasTitle && Adapt.course.get('_globals')._accessibility.altFeedbackTitle);
+    const title = (feedback.altTitle || feedback.title || this.get('title') || Adapt.course.get('_globals')._accessibility.altFeedbackTitle || '');
 
     const feedbackConfig = {
       isAltTitle,
-      altTitle,
       title: Handlebars.compile(title)(this.toJSON()),
       _classes: feedback._classes,
       ...(isLegacyConfig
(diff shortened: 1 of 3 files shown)


js/models/questionModel.js
```diff
@@ -284,10 +284,12 @@ class QuestionModel extends ComponentModel {
     };
 
     const altTitle = feedback.altTitle || Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
-    const title = feedback.title || this.get('title') || altTitle || '';
+    const hasTitle = Boolean(feedback.title || this.get('title'));
+    const isAltTitle = altTitle && !hasTitle;
+    const title = hasTitle ? (feedback.title || this.get('title')) : altTitle || '';
```
oliverfoster, line 289:
```suggestion
    const title = (feedback.title || this.get('title') || altTitle || '');
```

oliverfoster (changes requested):

schema/course.model.schema
```diff
@@ -361,7 +361,7 @@
               "type": "string",
               "title": "Alternative Feedback Title",
               "help": "Text read out by screen readers if no visual title is set for component question feedback.",
-              "default": "Feedback",
+              "default": "",
```
oliverfoster, line 364: I think this should probably stay to make sure there is always something in the absence of a given feedback title, component title or specified alt title. The notify dialog must have some specified accessible name.

swashbuck: Ok, that makes sense. I'll rework this.

kirsty-hames (approved):
This is working as described thanks 👍

js/models/questionModel.js
```diff
@@ -283,12 +283,12 @@ class QuestionModel extends ComponentModel {
       ) || {};
     };
 
-    const altTitle = feedback.altTitle || Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
-    const title = feedback.title || this.get('title') || altTitle || '';
+    const hasTitle = Boolean(feedback.title || this.get('title'));
+    const isAltTitle = Boolean(feedback.altTitle) || (!hasTitle && Adapt.course.get('_globals')._accessibility.altFeedbackTitle);
+    const title = (feedback.altTitle || feedback.title || this.get('title') || Adapt.course.get('_globals')._accessibility.altFeedbackTitle || '');
```
oliverfoster, line 288:
```suggestion
    const altFeedbackTitle = Adapt.course.get('_globals')._accessibility.altFeedbackTitle;
    const hasTitle = Boolean(feedback.title || this.get('title'));
    const isAltTitle = Boolean(feedback.altTitle) || (!hasTitle && altFeedbackTitle);
    const title = (feedback.altTitle || feedback.title || this.get('title') || altFeedbackTitle || '');
```

oliverfoster (changes requested):
👀 this seems to make sense. I think...

If an altTitle is set in the feedback object, then only the altTitle will be used as an aria label.
If no altTitle is set in the feedback object, then the course will visually display the feedback object title or the component title or use the global altFeedbackTitle as an aria label.

PR #552, opened 2024-06-10
Fix: Show static tooltips at end of event loop always

oliverfoster:
related https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/304

This allows safari to catch up on the initial position of the static tooltips, showing them in the correct place.

### Fix
* Show static tooltips at end of event loop always


kirsty-hames (approved):
I can confirm this resolves the [Safari issue](https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/304#pullrequestreview-2099274109) thanks @oliverfoster 👍

joe-allen-89 (approved):
👍

Issue #556, opened 2024-06-20
Component preview mode should be identified so extensions can behave in context


cahirodoherty-learningpool:
### Subject of the issue/enhancement/features
When running tests on content, components to be tested in isolation via the "preview" routing mode may still be interfered with by extensions that apply on data load etc. If we were able to declare context of preview mode then extensions could be configured to not fire accordingly

### Your environment
* FW [v5.39.15](https://github.com/adaptlearning/adapt_framework/releases/tag/v5.39.15)

PR #557, opened 2024-06-20
Update: Flag when routing isPreviewMode and reusing common functionality

cahirodoherty-learningpool:
Addresses #556 

### Update
* Flag when routing isPreviewMode and reusing common functionality

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -25,6 +25,7 @@ class Router extends Backbone.Router {
     this._navigationRoot = null;
     // Flag to indicate if the router has tried to redirect to the current location.
     this._isCircularNavigationInProgress = false;
+    this.isPreviewMode = false;
     this.showLoading();
     // Store #wrapper element and html to cache for later use.
     this.$wrapper = $('#wrapper');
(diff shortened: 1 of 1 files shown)


ethan-lp (approved):

js/router.js
```diff
@@ -80,47 +81,8 @@ class Router extends Backbone.Router {
   }
 
   handlePreview(...args) {
-    args = args.filter(v => v !== null);
```
lemmyadams, line 83: What was the purpose of all this code that is being deleted?

cahirodoherty-learningpool: All the deleted code is the same as what handleRoute already does, so I am just calling it to remove the duplication. The only exception being https://github.com/adaptlearning/adapt-contrib-core/pull/557/files#diff-2509a87ef563086fd634f97cc8372c3277e00655d6b9cbd7fc58c828ccdd0216R109-R111

lemmyadams (approved):

oliverfoster (approved):

PR #567, opened 2024-07-19
Fix: reset location info on changing language (fixes #566)

### LINKED ISSUE #566: Reset location info on changing language
In a course any two languages may be structurally different, so to avoid potentially invalid information the current location properties in location object should be cleared.

See also https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/issues/235

chris-steele:
#566 

### Fix
* reset location info on changing language


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #573, opened 2024-08-14
Fix: Emit router:location on contentObject reload

oliverfoster:
fixes https://github.com/adaptlearning/adapt-contrib-assessment/issues/204

### Fix
* Emit router:location on contentObject reload but not on language location reset


guywillis (approved):
👍

cahirodoherty-learningpool (approved):

PR #577, opened 2024-09-04
Update: jquery.resize.js to use ResizeObserver

### LINKED ISSUE #576: jquery.resize update to use ResizeObserver instead of manual calculations
### Subject of the issue
[ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) is now long fully supported https://caniuse.com/resizeobserver

It works in a similar way to MutationObserver, in that it allows well timed callbacks against the browser's rendering lifecycle, giving us the ability to immediately capture resize events as they happen and respond to them in a way that doesn't cause rendering glitches. It is much more performant and much less resource intensive than our current approach of measuring and comparing on expiring intervals and at strategic events.

oliverfoster:
fixes #576 

### Update
* Use browser API [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to reduce code complexity and increase performance

Refs:
* https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
* https://caniuse.com/resizeobserver

Note:
* Not ie11 compatible

### DIFF
--- a/js/views/contentObjectView.js
+++ b/js/views/contentObjectView.js
@@ -74,10 +74,11 @@ export default class ContentObjectView extends AdaptView {
         $(window).scrollTop(0);
       }
       Adapt.trigger(`${type}View:ready contentObjectView:ready view:ready`, this);
+      $.inview.unlock(`${type}View`);
       const styleOptions = { opacity: 1 };
       if (this.disableAnimation) {
         this.$el.css(styleOptions);
-        $.inview.unlock(`${type}View`);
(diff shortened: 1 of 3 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster:
Ready now, apologies

joe-allen-89 (approved):
Tested across multiple browsers, all working as expected 👍

PR #579, opened 2024-09-04
Update: inview.js to use IntersectionObserver

### LINKED ISSUE #578: inview update to use IntersectionObserver instead of manual calculations
### Subject of the issue
[IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) is now long fully supported https://caniuse.com/intersectionobserver

It works in a similar way to MutationObserver, in that it allows well timed callbacks against the browser's rendering lifecycle, giving us the ability to immediately capture onscreen events as they happen and respond to them in a way that doesn't cause rendering glitches. It is much more performant and much less resource intensive than our current approach of measuring and comparing on expiring intervals and at strategic events.

oliverfoster:
fixes #578

### Update
* Use browser API [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) to reduce code complexity and increase performance

Refs:
* https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver
* https://caniuse.com/intersectionobserver

Note:
* Not ie11 compatible

### DIFF
--- a/js/views/contentObjectView.js
+++ b/js/views/contentObjectView.js
@@ -74,11 +74,10 @@ export default class ContentObjectView extends AdaptView {
         $(window).scrollTop(0);
       }
       Adapt.trigger(`${type}View:ready contentObjectView:ready view:ready`, this);
-      $.inview.unlock(`${type}View`);
       const styleOptions = { opacity: 1 };
       if (this.disableAnimation) {
         this.$el.css(styleOptions);
-        $.inview();
+        $.inview.unlock(`${type}View`);
         _.defer(() => {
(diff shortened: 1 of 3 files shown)


oliverfoster:
Ready now, apologies

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
Tested with bookmarking on chrome, firefox and edge, all working as expected 👍

StuartNicholls (approved):

Issue #584, opened 2024-09-25
_isLocked set via _lockType is overridden by trickle


oliverfoster:
### Subject of the issue
We have a plugin that shows blocks in an article where the article has `_lockType: sequential` on its block models, the article also has trickle enabled at the article level. 

The block models are set to `_isLocked: true` initially:
https://github.com/adaptlearning/adapt-contrib-core/blob/e5a18ef4359cfbcf0cde433f6fea046779713ef4/js/models/adaptModel.js#L783

Trickle then overrides the block models when it comes to set the `_isLocked` attributes on the course descendants:
https://github.com/adaptlearning/adapt-contrib-trickle/blob/6a84eb56fef11116860c4fabdd0d41281498fd14/js/models.js#L239
```js
  // Apply only changed locking states
  Object.entries(locks).forEach(([ id, isModelLocked ]) => {
    const model = modelsById[id];
    const wasLocked = model.get('_isLocked');
    if (wasLocked === isModelLocked) return;
    model.set('_isLocked', isModelLocked);
  });
```

This can be fixed using the `lockedAttributes` for the `AdaptModel`, allowing trickle and core to force the model attribute to `_isLocked: true` until the same plugin unlocks it. Core essentially overrides trickle at the attribute level until core is ready to unlock the model.

oliverfoster:
completed in prs https://github.com/adaptlearning/adapt-contrib-trickle/pull/231 https://github.com/adaptlearning/adapt-contrib-core/pull/585

PR #585, opened 2024-09-25
New: _isLocked becomes a lockable attribute

oliverfoster:
part of #584 

### New
* _isLocked becomes a lockable attribute allowing plugins to force _isLocked: true

### Fix
* Do not lock when setting initial defaults (this can occur when cloning a model)
* Make sure to copy locked attribute states on ComponentModel.deepClone
* Clarified QuestionModel._canSubmit locking as model properties rather than an init function call

### Notes
* The core will softly warn to the console in the period between this change and changing other plugins (trickle) to provide their pluginName when setting _isLocked.

requires https://github.com/adaptlearning/adapt-contrib-trickle/pull/231

### DIFF
--- a/js/models/adaptModel.js
+++ b/js/models/adaptModel.js
@@ -49,6 +49,12 @@ export default class AdaptModel extends LockingModel {
     };
   }
 
+  lockedAttributes() {
+    return {
+      _isLocked: true
+    };
+  }
+
   /**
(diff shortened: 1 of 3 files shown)


swashbuck (approved):
👍

cahirodoherty-learningpool (approved):
👀

joe-replin (approved):

PR #593, opened 2024-10-16
New: Added toggleClass to AdaptModel and ItemModel

oliverfoster:
fixes #592 

### New
* `toggleModelClass(model, className, hasClass)` in `modelHelpers.js`
* `AdaptModel.prototype.toggleClass(className, hasClass)`
* `ItemModel.prototype.toggleClass(className, hasClass)`

PR #595, opened 2024-10-21
New: Add "has-error" class for use with question components (fixes #594)

### LINKED ISSUE #594: Add error class to components

Add an error class to question components when the user attempts to submit a component that is not yet ready to submit. This will allow individual question components to implement their own error styling.

Relates to:
https://github.com/adaptlearning/adapt-contrib-instructionError/issues/12 - In progress
https://github.com/adaptlearning/adapt-contrib-instructionError/issues/14

swashbuck:
Fix #594 

### New
* Adds the error class `has-error` to question components when the question cannot be submitted.

### Testing
1. Install and enable [Instruction Error](https://github.com/adaptlearning/adapt-contrib-instructionError) which will allow you to submit a question before it is completed (ex. before an option is selected in an MCQ)
2. Select the "Submit" button before making a choice
3. The component should now have the `has-error` class.
4. Next, make a selection for the question. If the question has multiple items (e.g. Matching), make a selection for every item.
5. Select the "Submit" button
6. The `has-error` class should be removed.

### DIFF
--- a/js/views/questionView.js
+++ b/js/views/questionView.js
@@ -244,8 +244,8 @@ class QuestionView extends ComponentView {
   }
 
   showInstructionError() {
-    Adapt.trigger('questionView:showInstructionError', this);
     this.model.toggleClass('has-error', true);
+    Adapt.trigger('questionView:showInstructionError', this);
   }
 
   // Blank method for question to fill out when the question cannot be submitted
--- a/js/views/questionView.js
+++ b/js/views/questionView.js
@@ -245,9 +245,6 @@ class QuestionView extends ComponentView {
 
   showInstructionError() {
     Adapt.trigger('questionView:showInstructionError', this);
-  }
-
-  addErrorClass() {
     this.model.toggleClass('has-error', true);
   }
 
--- a/js/views/questionView.js
+++ b/js/views/questionView.js
@@ -49,7 +49,6 @@ class QuestionView extends ComponentView {
   preRender() {
     this.listenTo(this.model, {
       'change:_isEnabled': this.onEnabledChanged,
-      'change:_isSubmitted': this.onSubmittedChanged,
       'question:refresh': this.refresh
     });
 
@@ -78,11 +77,6 @@ class QuestionView extends ComponentView {
       this.enableQuestion();
     }
 
-  }
-
-  onSubmittedChanged(model) {
-    // Remove error class, if it exists
-    model.toggleClass('has-error', false);
   }
 
   // Used by the question to disable the question during submit and complete stages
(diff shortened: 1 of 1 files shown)


kirsty-hames (approved):
Tested with #593 and works as expected thanks 👍

js/views/questionView.js
```diff
@@ -240,6 +246,10 @@ class QuestionView extends ComponentView {
     Adapt.trigger('questionView:showInstructionError', this);
   }
 
+  addErrorClass() {
+    this.model.toggleClass('has-error', true);
+  }
+
```
oliverfoster, line 252: A whole function is not needed for one line.
Can you add it into `showInstructionError`, before the event trigger please?

js/views/questionView.js
```diff
@@ -49,7 +49,7 @@ class QuestionView extends ComponentView {
   preRender() {
     // Setup listener for _isEnabled
     this.listenTo(this.model, 'change:_isEnabled', this.onEnabledChanged);
-
+    this.listenTo(this.model, 'change:_isSubmitted', this.onSubmittedChanged);
```
oliverfoster, line 52:
```js
    this.listenTo(this.model, {
      'change:_isEnabled': this.onEnabledChanged,
      'change:_isSubmitted': this.onSubmittedChanged,
      'question:refresh': this.refresh
    });
```

js/views/questionView.js
swashbuck: I created a separate function for this because showInstructionError() doesn't actually relate to adding a generic error class to a component. Perhaps showInstructionError() should be renamed showError()?

oliverfoster: Before the event trigger please. It should be in a fit state when the event is called such that the event handles have the `has-error` class. 
No, please don't change the function name, I think "instruction error" is a good description of the phase of the execution.

js/views/questionView.js
```diff
@@ -175,6 +181,7 @@ class QuestionView extends ComponentView {
 
     if (!canSubmit) {
       this.showInstructionError();
+      this.addErrorClass();
```
oliverfoster, line 184:
```suggestion
```

js/views/questionView.js
swashbuck: Done.

js/views/questionView.js
```diff
@@ -79,6 +80,11 @@ class QuestionView extends ComponentView {
 
   }
 
+  onSubmittedChanged(model) {
+    // Remove error class, if it exists
+    model.toggleClass('has-error', false);
```
oliverfoster, line 85: Could you move this to `onResetClicked`? Is there a reason why you need to do it when `_isSubmitted` changed?

swashbuck: Yep, that makes sense. I've made this change.

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #597, opened 2024-10-23
Fix: inview reprocess on window resize (fixes #596)

### LINKED ISSUE #596: Inview on resized window doesn't trigger correctly
### Subject of the issue
Inview on resized window doesn't trigger correctly.

oliverfoster:
fixes #596 

Inview was using old window dimensions to measure inview positions. These must be recalculated when the window is resized.

### Fix
* Trigger reprocess on window resize

### DIFF
--- a/libraries/inview.js
+++ b/libraries/inview.js
@@ -48,6 +48,7 @@
       process();
     }
   };
+  $(window).on('resize', wndw.resize);
   // handler functions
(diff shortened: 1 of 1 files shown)


swashbuck (approved):
👍

cahirodoherty-learningpool (approved):
👀

PR #599, opened 2024-10-23
Fix: inview return undefined when no element supplied (fixes #598)

### LINKED ISSUE #598: Inview throws error when no element passed
### Subject of the issue

`$([]).onscreen()` will error, it should return undefined. It is currently preventing bookmarking from returning users to their previous page due to this line https://github.com/adaptlearning/adapt-contrib-bookmarking/blob/f3c95c09426bfa330a687a36883974b20b31db03/js/adapt-contrib-bookmarking.js#L144

oliverfoster:
fixes #598 

### Fix
* Allow running `onscreen()` with no elements

### DIFF
--- a/libraries/inview.js
+++ b/libraries/inview.js
@@ -156,6 +156,7 @@
     return locks.includes(name);
   }
   function getMeasurement (element) {
+    if (!element?.getBoundingClientRect) return;
     const offset = element.getBoundingClientRect();
     const height = offset.height;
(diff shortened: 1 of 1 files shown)


swashbuck (approved):
👍

joe-replin (approved):

cahirodoherty-learningpool (approved):
👀

PR #601, opened 2024-10-29
Fix: CSS to animate popups, dialog for accessibility (fixes #600)

### LINKED ISSUE #600: Use css to animate popups and the dialog element for accessibility
### Subject of the issue
Using velocity to animate the drawer and notify is not great.
Using tabindex + aria-hidden to manage popup layers is not great.

We can now use css to animate and the dialog element to control popup layers.

oliverfoster:
fixes #600 

tbc 

### Fix
* A sentence describing each fix

### Update
* A sentence describing each update

### New
* A sentence describing each new feature

### Breaking
* A sentence describing each breaking change

### DIFF
--- a/js/views/drawerView.js
+++ b/js/views/drawerView.js
@@ -1,9 +1,18 @@
 import Adapt from 'core/js/adapt';
+import shadow from '../shadow';
 import a11y from 'core/js/a11y';
 import DrawerItemView from 'core/js/views/drawerItemView';
 import Backbone from 'backbone';
+import {
+  transitionNextFrame,
+  transitionsEnded
+} from '../transitions';
 
 class DrawerView extends Backbone.View {
+
+  tagName() {
+    return 'dialog';
+  }
 
   className() {
     return [
@@ -14,7 +23,6 @@ class DrawerView extends Backbone.View {
 
   attributes() {
     return {
-      role: 'dialog',
       'aria-modal': 'true',
       'aria-labelledby': 'drawer-heading',
       'aria-hidden': 'true',
@@ -32,6 +40,7 @@ class DrawerView extends Backbone.View {
   initialize() {
     this._isVisible = false;
     this.disableAnimation = Adapt.config.has('_disableAnimation') ? Adapt.config.get('_disableAnimation') : false;
+    this.$el.toggleClass('disable-animation', Boolean(this.disableAnimation));
     this._globalDrawerPosition = Adapt.config.get('_drawer')?._position ?? 'auto';
     this.drawerDuration = Adapt.config.get('_drawer')?._duration ?? 400;
     this.setupEventListeners();
@@ -41,6 +50,7 @@ class DrawerView extends Backbone.View {
   setupEventListeners() {
     this.onKeyUp = this.onKeyUp.bind(this);
     $(window).on('keyup', this.onKeyUp);
+    this.el.addEventListener('click', this.onShadowClicked.bind(this), { capture: true });
   }
 
   onKeyUp(event) {
@@ -49,18 +59,24 @@ class DrawerView extends Backbone.View {
     this.hideDrawer();
   }
 
+  onShadowClicked(event) {
+    const dialog = this.el;
+    const rect = dialog.getBoundingClientRect();
+    const isInDialog = (rect.top <= event.clientY && event.clientY <= rect.top + rect.height &&
+      rect.left <= event.clientX && event.clientX <= rect.left + rect.width);
+    if (isInDialog) return;
+    event.preventDefault();
+    this.hideDrawer();
+  }
+
   render() {
     const template = Handlebars.templates.drawer;
-    $(this.el).html(template({ _globals: Adapt.course.get('_globals') })).prependTo('body');
-    const shadowTemplate = Handlebars.templates.shadow;
-    $(shadowTemplate()).prependTo('body');
+    this.$el.html(template({ _globals: Adapt.course.get('_globals') }));
     _.defer(this.postRender.bind(this));
     return this;
   }
 
   postRender() {
-    this.$('a, button, input, select, textarea').attr('tabindex', -1);
-
     this.checkIfDrawerIsAvailable();
   }
 
@@ -74,7 +90,6 @@ class DrawerView extends Backbone.View {
       .removeClass(`is-position-${this.drawerPosition}`)
       .addClass(`is-position-${position}`);
     this.drawerPosition = position;
-    this.drawerAnimationDir = (position === 'auto') ? (isRTL ? 'left' : 'right') : position;
   }
 
   openCustomView(view, hasBackButton = true, position) {
@@ -109,7 +124,8 @@ class DrawerView extends Backbone.View {
     return (this._isVisible && this._isCustomViewVisible === false);
   }
 
-  showDrawer(emptyDrawer, position = null) {
+  async showDrawer(emptyDrawer, position = null) {
+    shadow.show();
     this.setDrawerPosition(position);
     this.$el
       .removeClass('u-display-none')
@@ -121,9 +137,6 @@ class DrawerView extends Backbone.View {
       a11y.scrollDisable('body');
       this._isVisible = true;
     }
-
-    // Sets tab index to 0 for all tabbable elements in Drawer
-    this.$('a, button, input, select, textarea').attr('tabindex', 0);
 
     if (emptyDrawer) {
       this.$('.drawer__back').addClass('u-display-none');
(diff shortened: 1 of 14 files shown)


kirsty-hames (commented):
This is working great thanks @oliverfoster. I've tested the tabbing and screen reader capability in comparison to the current functionality (master). Drawer dialog seems consistent, but I've noted a couple of differences for Notify dialog. Not sure if these are expected or need further input?

- Notify dialog popup title doesn't display visual focus anymore for Chrome or Safari (Mac). FF (Mac + Win) is intermittent. Chrome and Edge (Win) work as expected.

- For Edge (Win) and FF (Mac), when tabbing through Notify dialog popup, after the close btn, focus is directed to the browser UI. Focus does loop and return to the popup elements e.g. title > prev btn > next btn > close btn > browser UI > title > prev btn etc. For Chrome (Mac + Win) and Safari (Mac), focus is contained to the popup only.

js/views/drawerView.js
```diff
@@ -30,17 +38,21 @@ class DrawerView extends Backbone.View {
...
   initialize() {
+    _.bindAll(this, 'onShadowClicked');
     this._isVisible = false;
-    this.disableAnimation = Adapt.config.has('_disableAnimation') ? Adapt.config.get('_disableAnimation') : false;
+    this.disableAnimation = Adapt.config.get('_disableAnimation') ?? false;
+    this.$el.toggleClass('disable-animation', Boolean(this.disableAnimation));
     this._globalDrawerPosition = Adapt.config.get('_drawer')?._position ?? 'auto';
-    this.drawerDuration = Adapt.config.get('_drawer')?._duration ?? 400;
+    const drawerDuration = Adapt.config.get('_drawer')?._duration ?? 400;
+    document.documentElement.style.setProperty('--drawerDuration', `${drawerDuration}ms`);
```
swashbuck, line 47: @oliverfoster I added the `drawerDuration` as a CSS custom property. Would it make sense to do the same with Notify so that the duration is configurable via JSON? If so, where would that configuration go? There is no `_notify` property like there is with `_drawer` in _config.json_.

oliverfoster: Adding a notify property is fine. Are we sure we want these to be differentiated and not just relative to the @duration attribute that already exists?

swashbuck: The existing `@duration` property was too short, so the Notify and Drawer were animating in too quickly. 400ms is what was previously used with velocity.

swashbuck: Alternatively, we could remove the speed from the JSON config and add a new `@duration` in the [Vanilla theme](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/324b1fa92f7de903a2f45e136b3317721d75f682/less/_defaults/_animations.less#L3) for larger animations. I'm not sure how often devs customize the drawer speed via JSON, but it does seem like more of a theme concern.

oliverfoster: I'm not sure about adding it as a json property for core, I can't imagine it will be that frequently modified.

Nor about adding it as a CSS property, as that would suggest we'd want to override it at runtime, perhaps with different durations for various types of notify?

A theme extensible property seems sensible and a larger duration of 400ms seems sensible.

So perhaps a @duration-x2 or similar? @duration-longer? Any precedence for x2 or longer?

swashbuck: Yep, that sounds good. Since we have `@duration-progress` already in Vanilla (used for the PLP bar), we could create `@duration-notify` and `@duration-drawer` so that it is clear where they would be used.

Should we remove the configurable `_drawer._duration` property as well to be consistent?

oliverfoster: No, I can see why you did it now. Sorry.

I think add in the notify property and do them both as CSS variables - prefixed appropriately.

Yup. It was a good idea.

oliverfoster:
> Not sure if these are expected or need further input?
> 
> * Notify dialog popup title doesn't display visual focus anymore for Chrome or Safari (Mac). FF (Mac + Win) is intermittent. Chrome and Edge (Win) work as expected.

I have removed the focus first call to allow the browser to handle it through the dialog.showModal(( function. Adapt therefore no longer tries to focus on the title. 

> * For Edge (Win) and FF (Mac), when tabbing through Notify dialog popup, after the close btn, focus is directed to the browser UI. Focus does loop and return to the popup elements e.g. title > prev btn > next btn > close btn > browser UI > title > prev btn etc. For Chrome (Mac + Win) and Safari (Mac), focus is contained to the popup only.

I will make sure that the default browser behaviour is utilised for the loop around. We should not stop the focus from looping around into the browser UI if that is expect.

swashbuck:
I've removed the Drawer easing properties from the schemas in [4e00cad](https://github.com/adaptlearning/adapt-contrib-core/pull/601/commits/4e00cadcbcce6a4cc89c48d61e38a570766f8980). These reference easing strings that are specific to Velocity JS and are not available as a CSS [transition-timing-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-timing-function) value.

These easing methods could be reproduced using `cubic-bezier()` values, but we would want to set up new variables in the theme (ex. `@animation-easing-easeInOutQuart`). Currently, I've just switched all the easing functions to use `@animation-easing` in the [Vanilla PR](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/531).

oliverfoster:
> I've removed the Drawer easing properties from the schemas [..]
> These easing methods could be reproduced using `cubic-bezier()` values, [..] Not sure that it's worth the effort to support Drawer easing options via config?

On reflection, I'd like this to be as least breaking as possible. I have made a sub pr to reintroduce the easing as the suggested css cubic-bezier functions.

https://github.com/adaptlearning/adapt-contrib-core/pull/602

It is not possible to represent the bounce and elastic easing functions using just cubic-bezier functions, so I have gracefully replaced those with their nearest cubic-bezier counter parts, back and quint respectively.

References:
https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function
https://easings.net/#
https://joshcollinsworth.com/blog/easing-curves

oliverfoster:
👍

templates/shadow.hbs
```diff
@@ -1,2 +1 @@
-{{! drawerView.js will appends this on document's body}}
-<div id="shadow" class="shadow js-shadow u-display-none"></div>
+
```
swashbuck, line 1: @oliverfoster Can this template be removed completely? Is this for backwards compatibility?

swashbuck: I can't think of a use case, and we can always use pseudo selectors to add design elements via CSS. Definitely not something we have customized in the past. Not sure what anyone else thinks?

kirsty-hames: I've not amended the shadow template previously but I wonder if it would be useful to keep. With the dialog animation being updated to use CSS animate this gives us more flexibility to be creative whereas we've typically not touched this. Also, the notifyPopup template does include the [shadow](https://github.com/adaptlearning/adapt-contrib-core/blob/a1b6c57f8608d0a38ef6918ce3bc9bad6961f580/templates/notifyPopup.hbs#L106) so I think from a consistency prospective we should maintain both or remove both.

swashbuck: Thanks, @kirsty-hames . Let's leave it then.

kirsty-hames: Now the pressure is on to do something funky with it 💃

oliverfoster:
@kirsty-hames 

> I will make sure that the default browser behaviour is utilised for the loop around. We should not stop the focus from looping around into the browser UI if that is expect.

I've fixed it so that we support the existing `config.json:_accessibility._options_isPopupWrapFocusEnabled` behaviour. It now wraps when that property is `true` (default).

lemmyadams (approved):

kirsty-hames (approved):
Works great thanks 👍

oliverfoster:
Any other criticisms/suggestions on this? @swashbuck

js/a11y.js
```diff
@@ -438,13 +438,16 @@ class A11y extends Backbone.Controller {
     const config = this.config;
     $element = $($element).first();
 
+    const isInDOM = Boolean($element.parents('body').length);
+    if (!isInDOM) return false;
+
+    const isOutsideOpenPopup = this.isPopupOpen && !this.popupStack?.lastItem[0]?.contains($element[0]);
+    if (isOutsideOpenPopup) return null;
```
cahirodoherty-learningpool, line 445: Minor: Probably best to return Boolean false here

oliverfoster: This function intentionally returns `true`, `false` or `null` for where it should return a definitive response at the given element (`true` or `false`) or allow descending into the element's children (`null`).

The function `_findFirstForwardDescendant` can take `isReadable` as its selector. These lines are the important ones:

https://github.com/adaptlearning/adapt-contrib-core/blob/a1b6c57f8608d0a38ef6918ce3bc9bad6961f580/js/a11y.js#L666-L675

Where an element is outside of the open popup, the element may be an ancestor of the open popup and therefore its descendants should be searchable such that we can find the first `selector` element in the document.

Returning `null` is intended.

swashbuck (approved):
👍

oliverfoster:
I need to add the default animation to this before committing so that an absence of the corresponding vanilla code is less jarring.

oliverfoster:
Waiting for https://github.com/adaptlearning/adapt-contrib-vanilla/pull/531

PR #611, opened 2025-01-07
Fix: aria-label class not properly hidden (fixes #606)

### LINKED ISSUE #606: aria-label class not properly hidden

The `.aria-label` class is supposed to visually hide content while keeping it available to screen readers. But under certain conditions, that content is still:

1. Selectable, and visible when selected.
2. Blocking mouse clicks.

This seems to be specifically occurring on Windows, for example in Chrome. The issue is made worse if the user has set a minimum font-size in their browser settings, as that makes the hidden text bigger and more likely to have a negative effect.

### Your environment

* 5.41.12
* Windows, Chrome (doesn't occur on MacOS)

### Steps to reproduce

- Create a new course with `adapt create` and open it in Chrome on Windows
- Optionally, increase Chrome's minimum font size through Settings > Appearance > Customize fonts > Minimum Font size 
- Go to the 'Question Components' page
- Select all page content with mouse or <kbd>Ctrl + a</kbd>

### Expected behaviour

The visually hidden text should not be visible at any point 

### Actual behaviour

Visually hidden text is visble

### Screenshots (if you can)

The screenshot below shows a question component with all text content selected. This makes the following aria-label text visible:

* Completion status ("completed", `.js-a11y-completion-description`)
* Component type ("Multiple choice question")
* Feedback live region text, ("You answered incorrectly", `.js-btn-marking-label`). Since this text is positioned on top of the "show markings" button, it makes that part of the button unclickable with a mouse.

### Notes

It might be easiest replacing the current aria-label class:

https://github.com/adaptlearning/adapt-contrib-core/blob/89a0421ba06e127bfb0b5825a2bd7e839e06d49b/less/core/accessibility.less#L9-L25

with something more inline with other frameworks, such as [bootstrap's visually-hidden](https://github.com/twbs/bootstrap/blob/cbbb567b637e24882fd0fad62bca49e248afb9a4/scss/mixins/_visually-hidden.scss#L9-L16):

```css
.aria-label{
  display: block;
  width: 1px !important;
  height: 1px !important;
  padding: 0 !important;
  margin: -1px !important;
  overflow: hidden !important;
  clip: rect(0, 0, 0, 0) !important;
  white-space: nowrap !important;
  border: 0 !important;
}
```

This is a commonly used and well tested approach for visually hiding content, and the results of many years of the accessibility community trying out different approaches. It does not have the reported side effect.

hanshillen:
fixes #606

### Fix

* Updated `.aria-label` per discussion in #606

### DIFF
--- a/less/core/accessibility.less
+++ b/less/core/accessibility.less
@@ -3,25 +3,31 @@
 .component { position: relative; }
 .block { position: relative; }
 
-.aria-label, .visually-hidden {
-  display: block;
-  position: absolute !important;
-  width: 1px !important;
-  height: 1px !important;
-  padding: 0 !important;
-  margin: -1px !important;
-  overflow: hidden !important;
-  clip: rect(0, 0, 0, 0) !important;
-  white-space: nowrap !important;
-  border: 0 !important;
-  pointer-events: none;
+// Use to hide content visually while keeping it accessible to assistive technologies
+.visually-hidden() {  
+  display: block;  
+  position: absolute !important;  
+  width: 1px !important;  
+  height: 1px !important;  
+  padding: 0 !important;  
+  margin: -1px !important;  
+  overflow: hidden !important;  
+  clip: rect(0, 0, 0, 0) !important;  
+  white-space: nowrap !important;  
+  border: 0 !important;  
 }
 
-.aria-label-focusable, .visually-hidden-focusable {
- &:not(:focus):not(:focus-within) {
-    .visually-hidden
-  }
-}
+.aria-label {  
+  .visually-hidden();  
+  pointer-events: none;  
+} 
+
+// Use to only display content when it's focused
+.aria-label-focusable, .visually-hidden-focusable {  
+  &:not(:focus):not(:focus-within) {  
+     .visually-hidden;  
+   }  
+ } 
 
 // jquery.a11y start
 
--- a/less/core/accessibility.less
+++ b/less/core/accessibility.less
@@ -4,7 +4,7 @@
 .block { position: relative; }
 
 // Use to hide content visually while keeping it accessible to assistive technologies
-.visually-hidden() {  
+.visually-hidden {  
   display: block;  
   position: absolute !important;  
   width: 1px !important;  
--- a/less/core/accessibility.less
+++ b/less/core/accessibility.less
@@ -14,7 +14,7 @@
   clip: rect(0, 0, 0, 0) !important;
   white-space: nowrap !important;
   border: 0 !important;
-  display: none !important;
+  pointer-events: none;
 }
 
 .aria-label-focusable, .visually-hidden-focusable {
--- a/less/core/accessibility.less
+++ b/less/core/accessibility.less
@@ -3,32 +3,27 @@
 .component { position: relative; }
 .block { position: relative; }
 
-// jquery.a11y start
-// Invisible aria-labels
-// --------------------------------------------------
-.aria-label {
+.aria-label, .visually-hidden {
+  display: block;
   position: absolute !important;
-  &.relative {
-    position: relative !important;
-  }
-  left: 0 !important;
-  width: auto !important;
-  height: auto !important;
-  overflow: auto !important;
-  color: rgba(0,0,0,0) !important;
-  background: rgba(0,0,0,0) !important;
-  font-size: 1px !important;
+  width: 1px !important;
+  height: 1px !important;
   padding: 0 !important;
-  margin: 0 !important;
-  line-height: normal !important;
-  z-index: 1;
-}
-
-// Force aria-labels to hide on hidden
-// --------------------------------------------------
-.aria-label.aria-hidden {
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
   display: none !important;
 }
+
+.aria-label-focusable, .visually-hidden-focusable {
+ &:not(:focus):not(:focus-within) {
+    .visually-hidden
+  }
+}
+
+// jquery.a11y start
 
 // Hidden focus guard
 // --------------------------------------------------


less/core/accessibility.less
```diff
@@ -3,33 +3,28 @@
...
   padding: 0 !important;
-  margin: 0 !important;
-  line-height: normal !important;
-  z-index: 1;
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
+  display: none !important;
```
oliverfoster, line 17:
```suggestion
  pointer-events: none;
```
You have `display: block;` at the top and `display: none !important;` at the bottom.
Could you include `pointer-events: none;` such that the 1px x 1px space isn't clickable please?

oliverfoster (changes requested):

less/core/accessibility.less
hanshillen: Done

oliverfoster (approved):
👀 looks good

less/core/accessibility.less
```diff
@@ -3,33 +3,28 @@
 .component { position: relative; }
 .block { position: relative; }
```
kirsty-hames, line 5: Similar to the bootstrap example, is it worth adding a comment to instruct developers on its usage.

`// Hide content visually while keeping it accessible to assistive technologies`

less/core/accessibility.less
```diff
@@ -3,33 +3,28 @@
...
+  pointer-events: none;
 }
 
-// Force aria-labels to hide on hidden
-// --------------------------------------------------
-.aria-label.aria-hidden {
-  display: none !important;
+.aria-label-focusable, .visually-hidden-focusable {
+ &:not(:focus):not(:focus-within) {
+    .visually-hidden
```
kirsty-hames, line 22: `.visually-hidden` includes `pointer-events: none;` which we probably want to exclude if we intend to use this for clickable buttons e.g. "skip navigation". In which case we could utilise a mixin here. Please see suggestion below.

```
.visually-hidden() {
  display: block;
  position: absolute !important;
  width: 1px !important;
  height: 1px !important;
  padding: 0 !important;
  margin: -1px !important;
  overflow: hidden !important;
  clip: rect(0, 0, 0, 0) !important;
  white-space: nowrap !important;
  border: 0 !important;
}

.aria-label {
  .visually-hidden();
  pointer-events: none;
}

.aria-label-focusable, .visually-hidden-focusable {
 &:not(:focus):not(:focus-within) {
    .visually-hidden;
  }
}
```

kirsty-hames (commented):
This works as expected thanks @hanshillen but I've added some comments for your input please.

less/core/accessibility.less
hanshillen: Done

less/core/accessibility.less
```diff
@@ -3,33 +3,28 @@
...
-  margin: 0 !important;
-  line-height: normal !important;
-  z-index: 1;
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
+  pointer-events: none;
 }
```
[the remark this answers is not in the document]
hanshillen: Done

less/core/accessibility.less
hanshillen: Done

hanshillen:
I've addressed the requested changes.

less/core/accessibility.less
```diff
@@ -3,32 +3,33 @@
...
-  overflow: auto !important;
-  color: rgba(0,0,0,0) !important;
-  background: rgba(0,0,0,0) !important;
-  font-size: 1px !important;
-  padding: 0 !important;
-  margin: 0 !important;
-  line-height: normal !important;
-  z-index: 1;
+// Use to hide content visually while keeping it accessible to assistive technologies
+.visually-hidden() {
```
hanshillen, line 7: We should remove the parenthesis to keep the `.visually-hidden` alias available
```suggestion
.visually-hidden {  
```

kirsty-hames: Is there a specific use case to using `.visually-hidden` vs `.aria-label`? Same for `.aria-label-focusable` vs `.visually-hidden-focusable`.

kirsty-hames (approved):
Works as expected thanks 👍

oliverfoster (approved):

less/core/accessibility.less
```diff
@@ -3,32 +3,33 @@
...
-  display: none !important;
-}
+.aria-label {  
+  .visually-hidden();  
+  pointer-events: none;  
+} 
+
+// Use to only display content when it's focused
+.aria-label-focusable, .visually-hidden-focusable {  
+  &:not(:focus):not(:focus-within) {
```
kirsty-hames, line 27: Having revisited this, we should replace `:focus` with `:focus-visible`. This selector is useful to provide a different focus indicator based on the user's input modality (mouse vs. keyboard). Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
All other instances of `:focus` in Adapt have previously been [updated](https://github.com/adaptlearning/adapt-contrib-core/issues/231).

kirsty-hames (changes requested):

guywillis (approved):
👍

less/core/accessibility.less
oliverfoster: Would we not want elements with this mixin to always show if they've got focus, regardless of keyboard interactivity?

kirsty-hames: Yes, please ignore more comment above. This should remain as `:focus` (ref: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#focus_vs_focus-visible)

kirsty-hames (approved):

oliverfoster:
@cahirodoherty-learningpool would you mind +1ing please?

cahirodoherty-learningpool (approved):
👀

Issue #612, opened 2025-01-07
inview ignores locking on resize


oliverfoster:
### Subject of the issue
When fullscreening the video tag in adapt-contrib-media, the browser issues a resize event. inview ignores any inview locking on resize events allowing components to trigger inview completion upon resize.

PR #613, opened 2025-01-07
Fix: inview respecting locks on window resize (#612)

oliverfoster:
#612

### Fix
* Return early if trying to process inview handlers whilst locked


oliverfoster:
@swashbuck for medusa

swashbuck (approved):
👍

joe-allen-89 (approved):

PR #617, opened 2025-01-16
Fix: Popup support in a11y disabled (fixes #616)

### LINKED ISSUE #616: If accessibility is disabled dialogs cannot be opened
Observations:
 
- unopened dialogs are styled `display:none` by default
- [showModal ](https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/a11y/popup.js#L100) is not called if accessibility has been disabled 
- problem began with core v6.60.3

oliverfoster:
fixes #616 

### Fix
* Enable popup support when config.json:`_accessibility._isEnabled: false`

### DIFF
--- a/js/a11y/popup.js
+++ b/js/a11y/popup.js
@@ -89,16 +89,16 @@ export default class Popup extends Backbone.Controller {
    */
   _addPopupLayer($popupElement) {
     $popupElement = $($popupElement);
-    const config = this.a11y.config;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

chris-steele (approved):

swashbuck (approved):
👍

PR #621, opened 2025-02-03
Fix: Drawer close animation. Drawer and notify open focus

### LINKED ISSUE #619: Drawer closes prematurely when activating controls inside it by keyboard
When activating a button or a link inside the drawer using a keyboard, the drawer closes prematurely and the focused item does not get activated.

This is  caused by the fact that the `onShadowClicked` handler is set to handle the `click` event, which also fires for keyboard-activated links and buttons. 

https://github.com/adaptlearning/adapt-contrib-core/blob/2f7f5a72dc63576925af55b90d92a6df91cedf10/js/views/drawerView.js#L83

`onShadowClicked` uses the event's mouse coordinates to determine whether the click happened outside the dialog, Since key events have `clientX` and `clientY` set to `0`, the handler assumes the click was outside the drawer and closes it.

Easiest solution would be to instead listen for `mousedown` instead, or otherwise check first that the event is not a key event (e.g. by returning when clientX and ClientY are 0)

To reproduce:

Set up a course with adapt-contrib-resources. With the drawer open, tab to a filter button or resource item, and press Enter

### LINKED ISSUE #620: Drawer / Notify - Links in <dialog>
### Subject of the issue
If a `<dialog>` element, drawer or notify, contains a link, JAWS focuses on the link on opening instead of the start of the `<dialog>` element.

oliverfoster:
Additional fixes to pr https://github.com/adaptlearning/adapt-contrib-core/pull/601

fixes #619 
fixes #620 

### Fix
* Re-added the `focusFirst` calls removed in #601 to force the focus management of notify and drawer open, for issue #620
* Changed all escape key `keyup` handlers to `keydown` in order to better trap the even with screen readers
* Changed from `click` to `mousedown` for shadow click detection
* Changed the order of drawer close animation relative to the `dialog` element `close` function call, to fix #619
* Better support for deprecated [`drawer:closeDrawer` event used by pagelevelprogress](https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/blob/b9664407e0b1fd3a3bf75c069c006dbc0a215560/js/PageLevelProgressView.js#L41-L55) to ensure that the correct element is focused after pagelevel progress item click

### DIFF
--- a/js/deprecated.js
+++ b/js/deprecated.js
@@ -51,9 +51,9 @@ Adapt.on({
     logging.deprecated('Use drawer.remove, Adapt.trigger(\'drawer:remove\') will be removed in the future');
     drawer.remove();
   },
-  'drawer:closeDrawer'() {
+  'drawer:closeDrawer'($toElement) {
     logging.deprecated('Use drawer.close, Adapt.trigger(\'drawer:closeDrawer\') will be removed in the future');
-    drawer.close();
+    drawer.close($toElement);
   },
   'drawer:triggerCustomView'() {
     logging.deprecated('Use drawer.openCustomView(), Adapt.trigger(\'drawer:triggerCustomView\') will be removed in the future');
--- a/js/views/drawerView.js
+++ b/js/views/drawerView.js
@@ -78,12 +78,12 @@ class DrawerView extends Backbone.View {
   }
 
   setupEventListeners() {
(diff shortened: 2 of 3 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👍

PR #624, opened 2025-02-06
Fix: Prevent label clicks and keypresses from changing [aria-disabled] inputs (fixes #623)

### LINKED ISSUE #623: aria-disabled=true input still receives label click to trigger on change
If an `input` element has `[aria-disabled=true]` it is prevented from receiving click events by:

https://github.com/adaptlearning/adapt-contrib-core/blob/3cec6f13eec42d67c0fcfe388b9ce9593f7ceb84/js/a11y/browserFocus.js#L72-L76

However any associated `label[for]` is not prevented from receiving click events, those events propagate to the `input` and allow it to trigger a change event on the `input` in spite of its `[aria-disabled=true]` status.

This is a relatively new issue create by switching mcq and gmcq to `[aria-disabled=true]` from `[disabled]` in order to better facilitate screen readers at pr https://github.com/adaptlearning/adapt-contrib-mcq/pull/228

oliverfoster:
fixes #623 

### Fix
* Prevent label clicks and key presses from changing [aria-disabled] inputs
* Move aria-disabled handling into its own controller.

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -2,6 +2,7 @@ import Adapt from 'core/js/adapt';
 import offlineStorage from 'core/js/offlineStorage';
 import device from 'core/js/device';
 import location from 'core/js/location';
+import AriaDisabled from './a11y/ariaDisabled';
 import BrowserConfig from './a11y/browserConfig';
 import BrowserFocus from 'core/js/a11y/browserFocus';
 import FocusOptions from 'core/js/a11y/focusOptions';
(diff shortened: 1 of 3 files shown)


joe-replin (approved):
Tested in multiple builds including vanilla. Works as intended. Thank you so much.

cahirodoherty-learningpool (approved):
👀

kirsty-hames (commented):
Whilst this resolves the issue for label clicks, I'm still able to make a selection via keypress. Tested with voiceOver on Mac Safari.

oliverfoster:
@kirsty-hames Would you mind giving that a whirl please?

PR #627, opened 2025-02-13
Fix: Force body focus (fixes #626)

### LINKED ISSUE #626: Focus from menu to page
When using keyboard navigation only (no screen reader), the Tab focus will initially be on the 'Skip navigation' button when entering a topic, which means that the user can't immediately start scrolling with the up/down arrows - instead they have to select the button first.

oliverfoster:
fixes #626

When the router tries to focus on the `document.body` here:
https://github.com/adaptlearning/adapt-contrib-core/blob/230a2fa1ae12b6f1e854cb98964fc540286c55b7/js/router.js#L507

This pr forces the focus to the body element rather than trying to find the first focusable element - which had previously been the skip navigation button.

The outcome is that the body has focus making the window scrollable with the up/down arrows and then the first tab takes the learner to the skip navigation button, as before.

### Fix
* `a11y.focusFirst(document.body)` forces focus to body rather than first focusable element

### DIFF
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -730,6 +730,12 @@ class A11y extends Backbone.Controller {
   focusFirst($element, options) {
     options = new FocusOptions(options);
     $element = $($element).first();
+    const isBodyFocus = ($element[0] === document.body);
+    if (isBodyFocus) {
+      // force focus to the body, effectively starting the tab cursor from the top
+      this.focus($element, options);
+      return;
+    }
     if (this.isReadable($element)) {
       this.focus($element, options);
       return $element;


cahirodoherty-learningpool (approved):
👀

kirsty-hames (commented):
Thanks @oliverfoster. When using keyboard navigation only (no screen reader), this works as expected testing with Chrome, FF and Safari.

Running voiceOver in Chrome works as expected but not for Safari. The following issue is specific to using voiceOver keyboard controls <kbd>control</kbd> + <kbd>option</kbd> + <kbd>arrow</kbd> keys to navigate:
- On the menu, navigate to a menu item and select 'view' button.
- The page loads but focus remains on the menu item 'view' button.
- Using the same keyboard controls to navigate, the menu elements remain in focus (e.g. list item, menu item title, progress etc) even though you're on a page. See screenshots below.
- Only when using the <kbd>tab</kbd> key does the current page receive focus. 

**Expected behaviour seen in Chrome with voiceOver. On page load, the first body element receives focus (back button):**
chrome

**Unexpected behaviour seen in Safari with voiceOver. On page load, focus remain on the menu item 'view' button:**
safari

Tested running Safari Version 18.3 (20620.2.4.11.5), macOS Version 15.3 Sequoia. I do need to update to 15.3.1 Sequoia so I'll test this again when I do.

oliverfoster:
I'll let you have a crack at the new version before I have a go at this. That's very odd.

kirsty-hames:
> I'll let you have a crack at the new version before I have a go at this. That's very odd.

Hey @oliverfoster, I can confirm the [above issue](https://github.com/adaptlearning/adapt-contrib-core/pull/627#pullrequestreview-2629551779) still applies even after updating to macOS Version 15.3.1 Sequoia. And agreed, very odd :/

oliverfoster:
Safari doesn't allow focus to fall on the body element, even if it has any kind of permanent or dynamically added tabindex value or if any other element is blurred using `HTMLElement blur()`. This means that we can't use `document.body.focus()` to cause the screen reader to read the page afresh from the top as if it had navigated to a new website.

PR #633, opened 2025-02-28
Fix: trigger `inview` and `onscreen` events when element off screen

### LINKED ISSUE #630: `inview` and `onscreen` events not triggered when element off screen
With the changes applied in https://github.com/adaptlearning/adapt-contrib-core/releases/tag/v6.55.0, it seems that the `inview` and `onscreen` events no longer trigger when an element moves off screen. This means listeners such as https://github.com/adaptlearning/adapt-contrib-media/blob/e8b08fa5d5fe45ba2d0f95aa4675b9764c8e7dfd/js/mediaView.js#L310-L312 won't pause playback when off screen.

This seems to be caused by the addition of returning early if `!measurement.onscreen` when the library was rewritten:
https://github.com/adaptlearning/adapt-contrib-core/blob/4f7615bd175a0ee738aaa1a76f098f21689b6415/libraries/inview.js#L99

danielghost:
Fixes #630.

### Fix
* Trigger `inview` and `onscreen` events when element off screen

### DIFF
--- a/libraries/inview.js
+++ b/libraries/inview.js
@@ -96,7 +96,7 @@
       item.uniqueMeasurementId !== uniqueMeasurementId
     );
     item.uniqueMeasurementId = measurement.uniqueMeasurementId;
-    if (!hasMeasureChanged || !measurement.onscreen) return;
+    if (!hasMeasureChanged) return;
     switch (item.type) {
       case TYPE.onscreen:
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #636, opened 2025-03-10
Fix: add scroll padding to body to avoid focus obscured issue (fixes #635

### LINKED ISSUE #635: Focus obscured behind nav bar
If a focusable element receives focus while it's is positioned behind the navigation bar, it remains hidden.  

This is disorienting for sighted keyboard and screen magnifier users, and fails both WCAG SC 2.4.11
Focus Not Obscured (Minimum) and SC 2.4.12 Focus Not Obscured (Enhanced).

Note: this is specifically about elements receiving focus through tabbing and standard DOM `focus()` calls, not about moving focus using `router.navigateToElement()`. 

This can be solved by setting `scroll-padding-top`:

```less
body {
  scroll-padding-top: var(--adapt-navigation-height);
}
```

This will scroll any element into view if it receives focus while (partially) behind the nav bar.

hanshillen:
fixes #635

### Fix
* Add scroll padding to `body` to avoid focus from being obscured by navbar

### Testing
1. Press tab to move focus to a focusable element.
2. Scroll the page so that the next or previous (depending on `_isBottomOnTouchDevices`) focusable element is positioned behind the navigation bar (but within the viewport).
3. Press (shift) tab  to move focus to the obscured element
4. Confirm that the browser scrolls the focused element into view

Reference:
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top

### DIFF
--- a/less/core/accessibility.less
+++ b/less/core/accessibility.less
@@ -1,10 +1,9 @@
 // prevent elements from being obscured by nav bar when they receive focus
-body {
-  .is-nav-bottom & {
+:not(.adapt-scrolling), .adapt-scrolling body {
+  .is-nav-bottom& {
     scroll-padding-bottom: var(--adapt-navigation-height);
   }
-
-  .is-nav-top & {
+  .is-nav-top& {
     scroll-padding-top: var(--adapt-navigation-height);
   }
 }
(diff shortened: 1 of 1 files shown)


hanshillen:
@oliverfoster I have now, and have updated the selector accordingly. Let me know if I've missed anything else.

cahirodoherty-learningpool (approved):
👀

PR #645, opened 2025-03-25
New: Content object typegroup=group, router.addRouteToHistory (fixes #644)

### LINKED ISSUE #644: Boxmenu groups - navigation issue causing phantom sub-menu render
### Subject of the issue
If boxmenu is installed and a sub-menu contentobject is set to `_boxMenu._renderAsGroup: true`, then the router is pointed to navigate to that same contentobject, Adapt will render a phantom sub-menu instead of redirecting to the group on the main menu.

If a sub-menu contentobject is a group (as per box menu) and not meant to be rendered as sub-menu, then the plp will incorrectly navigate to the phantom sub-menu rendering of the group when a group is clicked in the plp. This is primarily because the router treats all contentobjects as the main contentobject and it cannot treat contentobjects as sub-parts of their parent. The only code which is currently aware that it should treat sub-menus as groups is the box menu which causes issues to arise in the plp and router.

related to https://github.com/adaptlearning/adapt-contrib-boxMenu/issues/212
related to https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/issues/247

oliverfoster:
fixes #644 

### New
* Added `model.isTypeGroup('group')` to tell the router that a contentobject is a sub-part of its parent and not a main contentobject
* Added `router.addRouteToHistory(hash)` to force the router to better track sub-contentobject navigation

### Fix
* Allow the router to navigate to sub-contentobject parts without re-rendering the contentobject
* The router now waits for the contentobject to become ready before assigning focus or navigating to a sub-part
* Renamed `router.isScrolling` to a more appropriate `router._isInNavigateTo`
* The router now adds sub-contentobject navigations to the url history

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -20,9 +20,11 @@ class Router extends Backbone.Router {
 
   initialize({ model }) {
     this.navigateToElement = this.navigateToElement.bind(this);
-    this._isBackward = false;
     this.model = model;
     this._navigationRoot = null;
+    this._isBackward = false;
+    this._isInNavigateTo = false;
+    this._shouldIgnoreNextRouteAction = false;
     // Flag to indicate if the router has tried to redirect to the current location.
     this._isCircularNavigationInProgress = false;
     this.isPreviewMode = false;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

lemmyadams (approved):

PR #647, opened 2025-03-25
Fix: Escape key closes only focused notify (fixes #646)

### LINKED ISSUE #646: Escape should only dismiss current popup
### Subject of the issue
Due to the way the onKeyDown listener is attached at the window, hitting escape will close all open notifypush and notifypopup views together.

oliverfoster:
fixes #646 

### Fix
* Escape key closes only focused notify


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #655, opened 2025-04-02
New: Focus at top of page when routing (fixes #626)

### LINKED ISSUE #626: Focus from menu to page

oliverfoster:
fixes #626 
a working alternative to #627 

Notifies the screen reader that a new content object is loaded, allows tabbing and arrowing forward onto the skip navigation button as well as scrolling with the arrow keys without a screen reader. Text is configurable at `_globals._accessibility._ariaLabels.topOfContentObject = '{{type}} {{displayTitle}}'`

`{{type}}` is `Main menu`, `Sub menu` or `Page`
`{{displayTitle}}` is the normalized `displayTitle`

Types are translatable at:
`_globals._accessibility._ariaLabels.course = "Main menu"`
`_globals._accessibility._ariaLabels.menu = "Sub menu"`
`_globals._accessibility._ariaLabels.page = "Page"`

### New
* Added topOfContentObject functionality which inserts an element at the top of the page, the element is updated and focused when routing

### Tested
I have done:

* voiceover with safari and chrome on macos 
* firefox and chrome and edge on windows with nvda and jaws
* android with chrome and talkback

All seem to read correctly, and it seems to make navigating the different content objects much more precise. There are the occasional double reads of `Page Presentation components` or `Main menu Adapt Version 5`  with jaws in firefox and edge, but most of the time they all behave nicely.

### DIFF
--- a/js/a11y/topOfPage.js
+++ b/js/a11y/topOfPage.js
@@ -35,7 +35,8 @@ export default class TopOfPage extends Backbone.Controller {
   goto() {
     const config = this.a11y.config;
     if (!config._isEnabled || !config._options._isPopupManagementEnabled) return;
-    this.a11y.focus(this.$element, { preventScroll: false, defer: true });
+    if (document.activeElement === this.$element[0]) return;
+    this.$element[0].focus();
   }
 
 }
--- a/js/a11y/topOfPage.js
+++ b/js/a11y/topOfPage.js
@@ -36,7 +36,7 @@ export default class TopOfPage extends Backbone.Controller {
     json.type = json._globals._accessibility._ariaLabels[json._type] || DEFAULT_TYPE_LABEL[json._type];
     json.heading = this.a11y.normalize(Handlebars.templates.heading(json));
     json.displayTitle = this.a11y.normalize(json.displayTitle);
-    const template = Handlebars.compile(json._globals._accessibility._ariaLabels.topOfPage || 'Top of {{type}} {{displayTitle}}');
+    const template = Handlebars.compile(json._globals._accessibility._ariaLabels.topOfPage || '{{type}} {{displayTitle}}');
     this.$element.html(template(json));
   }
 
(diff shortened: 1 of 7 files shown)


js/a11y/topOfPage.js
```diff
@@ -0,0 +1,50 @@
...
+
+  updateElement(location) {
+    const config = this.a11y.config;
+    if (!config._isEnabled || !config._options._isPopupManagementEnabled) return;
+    const json = location._currentModel.toJSON();
+    json._globals = Adapt.course.get('_globals');
+    json.type = json._globals._accessibility._ariaLabels[json._type] ?? DEFAULT_TYPE_LABEL[json._type];
+    json.heading = this.a11y.normalize(Handlebars.templates.heading(json));
+    json.displayTitle = this.a11y.normalize(json.displayTitle);
+    const template = Handlebars.compile(json._globals._accessibility._ariaLabels.topOfPage ?? '{{type}} {{displayTitle}}');
```
cahirodoherty-learningpool, line 39: Has this property been added (`_ariaLabels.topOfPage`)?

oliverfoster: Not yet. I was hoping for consensus on the solution before doing any more work. I need to do schemas and jsons and docs etc still.

kirsty-hames (approved):
This works great thanks @oliverfoster. I can replicate the occasional duplicate reading as you described but overall very consistent across browsers/OS.

I did a bit of research to see if there was a semantic element or ARIA we could use as an alternative to a custom `<div>` but found nothing appropriate.

VoiceOver Safari macOS and iPhone
VoiceOver Chrome and Firefox macOS

All browsers above without SR using keyboard only (Tab + arrow keys).

oliverfoster:
Cool, thanks for reviewing. I'll get the schema and example.json etc sorted.

cahirodoherty-learningpool (approved):
👀

kirsty-hames (approved):
👀

Issue #658, opened 2025-04-28
Improper use of aria-expanded for drawer dialog


kirsty-hames:
### Subject of the issue

The Drawer dialog modal shouldn't have an `aria-expanded` state. 

> Note: The presence of the aria-expanded attribute indicates control. Avoid including it on elements that do not control the expanded state of other elements.

See [MDN note](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded#:~:text=When%20the%20popup%20element%20is,should%20be%20set%20to%20true%20.&text=Note%3A%20The%20presence%20of%20the,expanded%20state%20of%20other%20elements) for reference.

For example, in the Accordion component, [`aria-expanded`](https://github.com/adaptlearning/adapt-contrib-accordion/blob/d98466c847543457efcaff0289d25f9bbfc85323/templates/accordion.jsx#L53) is applied to the `<button>` that controls the expand/collapse of the item content. This is the intended use.

This might just be legacy code prior to the Drawer changed to use `<dialog>` as `aria-expanded` isn't applied to the Notify `<dialog>`.

See [PR where <dialog> was introduced](https://github.com/adaptlearning/adapt-contrib-core/pull/601) for reference.

### Expected behaviour
`aria-expanded` should be removed from drawer `<dialog>` as this serves no purpose and it is not the intended use.

PR #663, opened 2025-05-12
Update: Replace dialog button aria-expanded with aria-haspopup dialog (fixes #657)

### LINKED ISSUE #657: Improper use of aria-expanded for nav buttons which trigger drawer / notify dialog
### Subject of the issue
The following issue has been reported by an external accessibility auditor:

> The aria-expanded="false" attribute is specified to the "Open course resources.", "Page progress" and "Visual accessibility settings" image buttons. However the mentioned image buttons displays a modal dialog on activating them and there is no expand/collapse functionality is associated with the mentioned image buttons. This might create confusion for screen reader users.

Fails WCAG Success Criteria 4.1.2 https://www.w3.org/WAI/WCAG20/Understanding/name-role-value

Having done some research I'm inclined to agree. Please see my points below but I'm interested in hearing thought from others.
- The use of `aria-expanded` assumes the button toggles the collapse as well as the expanding of the dialog. The button only triggers the expanding of the dialog. The dialog is self closing (either shadow or close button).
- The dialog isn't a sibling or child of the button.
- Whilst [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded) references the use of `aria-expanded` for `dialog` elements, I think this is in reference to non-modal dialogs (where focus can move outside of the dialog whilst still visible). Modal dialogs trap focus within the popup only and prevent learners from interacting with the rest of the page until the dialog is closed. Therefor the change in `aria-expanded` state isn't accessible, it will only be read as "collapsed".
- The [WCAG dialog (modal) example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/) doesn't include `aria-expanded`.

Some of these [points were made previously](https://github.com/adaptlearning/adapt-contrib-core/issues/103#issuecomment-1113003211) when the [initial issue was raised for WCAG fail of Drawer](https://github.com/adaptlearning/adapt-contrib-core/issues/103).

Reference: [initial issue raise to implement `aria-expanded`](https://github.com/adaptlearning/adapt-contrib-core/issues/409).

### Steps to reproduce
1. Using screen reader, navigate to the mentioned image buttons.
2. Listen that the incorrect state is announced.

### Expected behaviour
Whilst I don't think `aria-expanded` is the appropriate use, I do think it's useful to indicate the buttons do trigger the change in page content.  
As an alternative to `aria-expanded`, I wonder whether adding "Open" to the existing button `aria-labels` is sufficient? Like we already have for Resources, "Open course resources". However if people are in favour of the current implementation I'm happy to leave as is.

kirsty-hames:
### Update
- For navigation buttons that trigger a dialog to open (Notify popup or Drawer), the button `aria-expanded` has been replaced with `aria-haspopup="dialog"`.  Please refer to the [issue](https://github.com/adaptlearning/adapt-contrib-core/issues/657) for the research and discussion supporting this change. 
Fixes https://github.com/adaptlearning/adapt-contrib-core/issues/657

- Remove `aria-expanded` from Drawer `<dialog>` as this serves no purpose and it is not the intended use. See [MDN note](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded#:~:text=When%20the%20popup%20element%20is,should%20be%20set%20to%20true%20.&text=Note%3A%20The%20presence%20of%20the,expanded%20state%20of%20other%20elements) for reference below or refer to the [issue](https://github.com/adaptlearning/adapt-contrib-core/issues/658) description.
Fixes https://github.com/adaptlearning/adapt-contrib-core/issues/658

### Testing
Navigate to the Resources button in the nav bar using a screen reader. Depending on the browser/screen reader used, expect the following reading....

VoiceOver “Open course resources, dialog pop-up, button”
NVDA “Open course resources, button opens dialog”
JAWS “Open course resources, button has popup dialog”

VoiceOver Safari macOS and iPhone
VoiceOver Chrome and Firefox macOS

### Related PRs
Other navigation plugins that include `aria-expanded` for buttons that open Drawer or Notify popup dialogs will also need updating. Please see summary below.

- [Resources](https://github.com/adaptlearning/adapt-contrib-resources) and [Glossary](https://github.com/adaptlearning/adapt-contrib-glossary) have been addressed with this PR.

- PLP PR - https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/251

- Visua11y PR- https://github.com/cgkineo/adapt-visua11y/pull/130

### DIFF
--- a/js/views/drawerView.js
+++ b/js/views/drawerView.js
@@ -26,8 +26,7 @@ class DrawerView extends Backbone.View {
     return {
       'aria-modal': 'true',
       'aria-labelledby': 'drawer-heading',
-      'aria-hidden': 'true',
-      'aria-expanded': 'false'
+      'aria-hidden': 'true'
     };
   }
 
@@ -159,8 +158,7 @@ class DrawerView extends Backbone.View {
     this.setDrawerPosition(position);
     this.$el
       .removeClass('u-display-none')
-      .attr('aria-hidden', 'false')
-      .attr('aria-expanded', 'true');
+      .attr('aria-hidden', 'false');
     // Only trigger popup:opened if drawer is visible, pass popup manager drawer element
     if (!this._isVisible) {
       a11y.popupOpened(this.$el);
@@ -192,7 +190,6 @@ class DrawerView extends Backbone.View {
     }
 
     $('.js-drawer-holder').scrollTop(0);
-    $('.js-nav-drawer-btn').attr('aria-expanded', true);
     Adapt.trigger('drawer:opened');
 
     this.$el.addClass('anim-show-before');
@@ -236,7 +233,6 @@ class DrawerView extends Backbone.View {
     if (!force) await transitionsEnded(this.$el);
 
     this._customView = null;
-    $('.js-nav-drawer-btn').attr('aria-expanded', false);
     Adapt.trigger('drawer:closed');
 
     this._isVisible = false;
@@ -248,8 +244,7 @@ class DrawerView extends Backbone.View {
     this.$el
       .removeAttr('style')
       .addClass('u-display-none')
-      .attr('aria-hidden', 'true')
-      .attr('aria-expanded', 'false');
+      .attr('aria-hidden', 'true');
     this.setDrawerPosition(this._globalDrawerPosition);
   }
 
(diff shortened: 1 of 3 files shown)


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

Issue #656, opened 2025-04-09
Add unique and descriptive labels for question buttons


kirsty-hames:
### Subject of the issue
The following issue has been reported by an external accessibility auditor:

> Identical label “Show correct answer”, "Show your answer", "Show feedback" and "Submit" are specified for the buttons.
> As a result, screen reader users might not understand the purpose of the buttons effectively.

![Image](https://github.com/user-attachments/assets/c80eb9a4-04f3-4652-9758-41114752ecb4)

### Steps to reproduce
1. Using screen reader, navigate to the mentioned buttons.
2. Listen that the buttons label are identical.

### Expected behaviour
The screen reader should announce unique and descriptive labels for the buttons. 

Recommendation provided by the auditor:

> Provide unique and descriptive labels for the buttons using `aria-label` attribute.
> 
> Alternatively apply the following changes:
> 
> · Remove `aria-label` attribute from the source code of mentioned buttons.
> 
> · Specify `aria-labelledby` attribute to `<button>` element containing the mentioned buttons.
> 
> · Specify `id` attribute with unique value to `<button>` element containing mentioned buttons such as "Show feedback" and `<div>` element containing the heading text such as "Multiple Choice Question".
> 
> · Reference both `id` attribute values of `<button>` and `<div>` elements to `aria-labelledby` attribute respectively.

Based on the above solution, I'd suggest either adding the component `title` to the button `aria-label`. Or, use `aria-labelledby` as suggested to reference the component `title`.

oliverfoster:
Perhaps you could use a similar technique to the one used by mcq to group the items, but use it to group the buttons instead?

```jsx
<div 
  role="group"
  aria-labelledby={ariaQuestion ? null : (displayTitle || body || instruction) && `${_id}-header`}
  aria-label={ariaQuestion || null}
>
  <div>buttons stuff</div>
</div>
```

https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA17

PR #666, opened 2025-05-21
Add unique and descriptive labels for question buttons

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-core/issues/656

### Update
* `aria-describedby` added to question submit and feedback button to provide context of the question.

### Testing
Navigate to a question component using a screen reader. Depending on the browser/screen reader used, the submit and feedback button should read as “Submit, question title, button” and “Show feedback, question title, button” or similar.

VoiceOver Safari and Chrome macOS

### DIFF
--- a/templates/buttons.hbs
+++ b/templates/buttons.hbs
@@ -8,13 +8,13 @@
   <div class="js-answer-live-region aria-label" aria-live="assertive" aria-atomic="true"></div>
   {{/if}}
 
-  <button class="btn-text btn__action js-btn-action" aria-label="{{_buttons._submit.ariaLabel}}">
+  <button class="btn-text btn__action js-btn-action" aria-label="{{_buttons._submit.ariaLabel}}" aria-describedby="{{_id}}-heading">
     {{{_buttons._submit.buttonText}}}
   </button>
   
   <span class="js-btn-marking-label aria-label u-display-none">{{#all _isInteractionComplete _canShowMarking}}{{#if _isCorrect}}{{@root/_globals._accessibility._ariaLabels.answeredCorrectly}}{{else}}{{@root/_globals._accessibility._ariaLabels.answeredIncorrectly}}{{/if}}{{/all}}</span>
 
-  <button class="btn-text btn__feedback js-btn-feedback is-disabled" aria-label="{{_buttons._showFeedback.ariaLabel}}" aria-disabled="true">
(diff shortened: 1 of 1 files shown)


lemmyadams (approved):

oliverfoster:
Should this be more like the mcq version? https://github.com/adaptlearning/adapt-contrib-mcq/blob/a976b88f280af002e5b59de7a7f9bf3b8e0d5473/templates/mcq.jsx#L46

kirsty-hames:
> Should this be more like the mcq version?
> 
> https://github.com/adaptlearning/adapt-contrib-mcq/blob/a976b88f280af002e5b59de7a7f9bf3b8e0d5473/templates/mcq.jsx#L46-L47

I _think_ the use case is slightly different here but happy to take a suggestion if I've overlooked something.

For MCQ, either `aria-label` or `aria-labelledby` are used, not both. Dependant on whether `ariaQuestion` has been provided.

For question buttons, we want to use the existing button label and provide additional context. 

`aria-label` can only contain strings, whilst `aria-labelledby` and `aria-describedby` list `id`s.

`aria-labelledby` takes precedence over all other methods of providing an accessible name, including `aria-label`. Whilst `aria-describedby` provides additional context to the existing `aria-label`.

References:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby

oliverfoster:
The technique applied in the mcq is ARIA17 https://www.w3.org/TR/WCAG20-TECHS/ARIA17.html, it is the combination of `role=group` with `aria-label` or `aria-labelledby`.

The other part of it, the `ariaQuestion` value, was to override `aria-labelledby={{id}}-heading` and replace it with `aria-label={{ariaQuestion}}` where the contents of `{{id}}-heading` were not descriptive or specific enough.

It should be possible to wrap the button with a `role=group` and the relevant label, and use `ariaQuestion`.

That would mean that the question widget controls and the question buttons would have the same grouping label. With this pr as it is, they would not have the same grouping / descriptive label.

kirsty-hames:
This sounds great thanks @oliverfoster - I'll give it a go...

oliverfoster (approved):

PR #671, opened 2025-06-20
New: Add _ariaLevel to core schemas (fix #625)

### LINKED ISSUE #625: AAT: Add ariaLevel to article/block/component schemas
## Subject of the enhancement
Currently, you can override an element's title by adding a custom `_ariaLevel` value.

```
"_ariaLevel": 5,
```

However, this feature is not available in the authoring tool.

## Suggestion
Add the ability to set `_ariaLevel` values in the AAT. It looks like having it default to 0 would mean that it's basically ignored and inherits the global settings.

## Use case in AAT
One of our projects recently had an accessibility audit. The course pages were set up using sections where each section heading was a full width Text component.

- Text Component - "Section 1 heading"
- Component
- Component
- Text Component - "Section 2 heading"
- Component
- Component

The consultants flagged that the section heading components' titles had the same heading levels as the component titles that "lived" in the section. The recommendation was to adjust the headings of each section component. So, if the section heading components had an `aria-level` of 3, the components that live in each section should use an `aria-level` of 4.

Since we do not have access to `_ariaLevel` overrides in the AAT, the solution was to create new articles for every section and move the section heading text there. This fixes the issue but is a time-consuming change to make.

swashbuck:
Fixes #625

### New
* Adds `_ariaLevel` to core schemas
* Adds required migration scripts

### DIFF
--- a/migrations/v6.js
+++ b/migrations/v6.js
@@ -654,3 +654,50 @@ describe('core - update to v6.60.3', async () => {
     content: [{ _type: 'course' }]
   });
 });
+
+describe('core - @@CURRENT_VERSION to @@RELEASE_VERSION', async () => {
+  let contentModels;
+
+  whereFromPlugin('core - from @@CURRENT_VERSION', { name: 'adapt-contrib-core', version: '<@@RELEASE_VERSION' });
+
+  whereContent('core - where content models', async (content) => {
+    const acceptedTypes = ['article', 'page', 'menu', 'block', 'component'];
+    contentModels = content.filter(obj => acceptedTypes.includes(obj._type));
+    return contentModels.length;
+  });
+
+  mutateContent('core - add _ariaLevel attribute', async (content) => {
+    contentModels.forEach(model => {
+      _.set(model, '_ariaLevel', 0);
+    });
+    return true;
+  });
+
+  checkContent('core - check _ariaLevel added', async (content) => {
+    const isValid = contentModels.every(model => _.has(model, '_ariaLevel') && model._ariaLevel === 0);
+    if (!isValid) throw new Error('core - _ariaLevel not added to all content models');
+    return true;
+  });
+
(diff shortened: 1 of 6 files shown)


oliverfoster:
Could you explain the use-cases where automatically calculations didn't work please?

swashbuck:
> Could you explain the use-cases where automatically calculated levels didn't work please?

If our courses had _initially_ been set up to use a structure that reflected the content, then the automatic levels would have worked. The problem relates to reworking a large number of courses where using `_ariaLevel` overrides would have saved us more development time. More info in the #625 ticket description.

kirsty-hames:
> > Could you explain the use-cases where automatically calculated levels didn't work please?
> 
> If our courses had _initially_ been set up to use a structure that reflected the content, then the automatic levels would have worked. The problem relates to reworking a large number of courses where using `_ariaLevel` overrides would have saved us more development time. More info in the #625 ticket description.

Whilst `_ariaLevel` overrides are redundant if using the CO/A/B/C structure I agree it's useful to have the option to override for instances that haven't followed this structure. 

Automated heading levels makes light work and reduces human error in labelling appropriate heading levels. Do we need some form of instruction/documentation to guide authors in using the structure when authoring courses? And note `_ariaLevel` should remain an override for existing courses?

swashbuck:
Ready for review.

@kirsty-hames Let me know if this updated instruction text makes sense. I tried to incorporate some of the wording you used.

> The aria-level attribute of the title used to determine the heading level. When set to 0 (default), the level will be automatically determined by the element type. Only override if content does not properly use the CO/A/B/C structure.

kirsty-hames (approved):
The schemas look good to me thanks @swashbuck. Has this been tested in AAT?

swashbuck:
> The schemas look good to me thanks @swashbuck. Has this been tested in AAT?

Thanks, @kirsty-hames . It has _not_ actually been tested in any AAT instances.

joe-allen-89 (approved):
👍

PR #673, opened 2025-06-25
New: Added notify.read(body) (fixes #672)

### LINKED ISSUE #672: Add easy api for screen reader only notifications
Currently the only way to have the screen reader read is to run:
```js
notify.push({
  _type: 'a11y-push',
  body: 'Loading.'
});
```

This doesn't allow for a delay, in the case of needing to wait for focusing to finish, or facilitate a wait  until the read timeout has expired.

oliverfoster:
fixes #672 

Primarily adds `await notify.read(body)`, a mechanism through which to instruct the screen reader to read. It waits for 200ms after the last focus event before reading, and for 200ms per word, plus 150ms for the text to read.

### Fix
* Reduced nesting in the `notifyPushCollection`
* `NotifyPushView` switched from `HTMLDialogElement.show()` to `HTMLDialogElement.open = true` to stop the browser from scrolling when opening a push notification 

### New
* `get a11y.timeSinceLastFocus` returns the ms since the last focus event
* `notify.create(createObject._delay)` to allow for a delay before showing a push notification
* `NotifyModel` has had a `close` and `await onClosed()` function to allow close signaling
* Return the `notifyModel` instance from `notify.create()` for push type events, this allows for access to the `await notifyModel.onClosed()` on the push notification
* `await notify.read(string)` to trigger an `a11y-push` notification, with a 200ms delay since the last focus event, give 200ms per word + 150ms to read, and resolve on the close the notify

### DIFF
--- a/js/views/notifyPushView.js
+++ b/js/views/notifyPushView.js
@@ -66,7 +66,7 @@ export default class NotifyPushView extends Backbone.View {
   render() {
     const data = this.model.toJSON();
     const template = Handlebars.templates.notifyPush;
-    this.$el.html(template(data)).appendTo('.notify__push-container');
+    this.$el.html(template(data));
 
     _.defer(this.postRender.bind(this));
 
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

PR #677, opened 2025-06-26
New: Added popup:closing event + wait queue, allowed a11y.setPopupClose for notify (fixes #676)

### LINKED ISSUE #676: Cannot change focal element on notify close
### Subject of the issue
When opening a notify, it captures the `document.activeElement` to return there upon closing. There is no easy way to change the element that should be focused after the notify closes. 

This feature was introduced in 2018 in [this commit](https://github.com/adaptlearning/adapt-contrib-core/commit/244af2a2ff56a840cf463e982f1e2d6b5f4d8eac#diff-d5d956ed97fb3ba2c05b53c661af8cc71a1d68646cdb4be37d58e3ecbcde4565R141), with the following lines in play:

https://github.com/adaptlearning/adapt-contrib-core/blob/244af2a2ff56a840cf463e982f1e2d6b5f4d8eac/js/views/notifyView.js#L141
https://github.com/adaptlearning/adapt-contrib-core/blob/244af2a2ff56a840cf463e982f1e2d6b5f4d8eac/js/views/notifyView.js#L228
https://github.com/adaptlearning/adapt-contrib-core/blob/73ef2176dae0750f99ab6e3fb5578ef8dfc85d69/js/popupManager.js#L14-L22

And can be seen in today's code at these lines:

https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/views/notifyPopupView.js#L130
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/views/notifyPopupView.js#L206
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L152-L158

It would be ideal if we could use the `a11y.setPopupCloseTo` for the notify:

https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y.js#L909-L918
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L230-L233

oliverfoster:
fixes #676 

Allowed `a11y.setPopupCloseTo` to work with the notify. Added the `popup:closing` event, followed immediately by an `await wait.queue()` to allow other plugins to calculate and assign an element to `a11y.setPopupCloseTo($element)`.

This works principally because the notify does not need to retain the `document.activeElement` when opening, as the `popup` API already performs this function at the following lines:

https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L78
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L93
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L153
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L176
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/a11y/popup.js#L158

### Fix
* Removed code from notify to return to the previous active element and rely on the popup API to handle it

### New
* Added `popup:closing` event with an `await wait.queue` such that other plugins can listen to the event and `wait.begin()`, `a11y.setPopupCloseTo($element)` and `wait.end()`

### DIFF
--- a/js/a11y/popup.js
+++ b/js/a11y/popup.js
@@ -1,5 +1,6 @@
 import Adapt from 'core/js/adapt';
 import logging from '../logging';
+import wait from 'core/js/wait';
 
 /**
  * Tabindex and aria-hidden manager for popups.


joe-allen-89 (changes requested):
Currently throwing an error, missing import for wait

lemmyadams (approved):

PR #679, opened 2025-06-26
New: Added option returnNewDescendants to AdaptView addChildren and addDescendants (fixes #678)

### LINKED ISSUE #678: Not easily able to derive the next set of dynamically added descendants
### Subject of the issue
When calling `addChildren` on an `AdaptView`, there is no easy way to fetch the newly added views. This is as `addChildren` only returns the number of descendants added and not a list of the descendants.

https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/views/adaptView.js#L144-L199

oliverfoster:
fixes #678 

### New
* Added option `returnNewDescendants` to `AdaptView` `addChildren` and `addDescendants` to allow other plugins to see which new views will be added


cahirodoherty-learningpool (approved):
👀

PR #697, opened 2025-09-30
Fix: Manage focus to prevent jaws virtual cursor from scrolling (#695)

### LINKED ISSUE #695: jaws click on background scrolls to top
### Subject of the issue
With JAWS active and when focused on a form input element, clicking any non-interactive element will cause the focused element to blur and the virtual cursor of JAWS to move to the body element, which then scrolls the page to the top.

oliverfoster:
fixes #695 

### Fix
* Ensure safari label+input label click correctly focuses on input

### Update
* Refocus on the existing active element when a non-interactive element is clicked

### New
* Added `a11y.currentActiveElement`

### DIFF
--- a/js/a11y/browserFocus.js
+++ b/js/a11y/browserFocus.js
@@ -74,6 +74,9 @@ export default class BrowserFocus extends Backbone.Controller {
     const $stack = $([...$element.toArray(), ...$element.parents().toArray()]);
     const $focusable = $stack.filter(config._options._tabbableElements);
     if (!$focusable.length) {
+      const element = this.a11y.currentActiveElement;
+      // refocus on the existing active element to stop jaws from scrolling
+      element?.focus();
       return;
     }
     const $closestFocusable = $element.closest(config._options._tabbableElements);
--- a/js/a11y.js
+++ b/js/a11y.js
@@ -185,6 +185,10 @@ class A11y extends Backbone.Controller {
     return Date.now() - this._lastFocusTime;
   }
 
+  get currentActiveElement() {
+    return this._activeElements[0];
+  }
+
   get previousActiveElement() {
     return this._activeElements[1];
   }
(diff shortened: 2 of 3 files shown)


js/a11y/browserFocus.js
```diff
@@ -45,19 +45,31 @@ export default class BrowserFocus extends Backbone.Controller {
     if (event.target !== event.currentTarget) {
       return;
     }
-    // Do not auto next if the focus isn't returning to the body
-    if (!$(event.relatedTarget).is('body')) {
+    // Do not auto next if the focus isn't returning to the body or html element
+    if (!$(event.relatedTarget).is('body, html')) {
       return;
     }
```
danielghost: It looks like the same issue occurs after submitting a question and then clicking on one of the options again. At that point these conditions are met. Is it acceptable to move these conditions to execute after https://github.com/adaptlearning/adapt-contrib-core/pull/697/files#diff-3ff4169174c1fed2be880d3992f0884ac57b087ee5195aad74c799717f73a8b8R55-R61 rather than before?

ethan-lp (approved):

js/a11y/browserFocus.js
```diff
@@ -45,19 +45,32 @@ export default class BrowserFocus extends Backbone.Controller {
...
+    // or if we're not losing focus
+    if (!$(event.relatedTarget).is('body, html') && event.relatedTarget !== null) {
       return;
     }
     // Check if element losing focus is losing focus
-    // due to the addition of a disabled class
-    if (!$element.is('[disabled]') && $element.css('display') !== 'none' && $element.css('visibility') !== 'hidden') {
+    // due to the addition of a disabled class, display none, visibility hidden,
+    // or because it has been removed from the dom
+    if (!$element.is('[disabled]') && $element.css('display') !== 'none' && $element.css('visibility') !== 'hidden' && $element.parents('html').length) {
```
swashbuck: This line is really long and complex. Suggest using more [descriptive conditions](https://github.com/adaptlearning/documentation/blob/master/01_cross_workstream/style_guide.md#use-descriptive-conditions).

PR #706, opened 2025-10-14
Fix: TextInput edge case where _canSubmit early false

oliverfoster:
fixes https://github.com/adaptlearning/adapt-contrib-textInput/issues/159

### Fix
* Check that button is in SUBMIT state with _canSubmit = false before disabling


ethan-lp (approved):

lemmyadams (approved):

taylortom (approved):
Looks good to me 👀

Issue #705, opened 2025-10-13
Start location ignored for reset courses on language change


danielghost:
If a course is reset with [_restoreStateOnLanguageChange=true](https://github.com/adaptlearning/adapt-contrib-languagePicker?tab=readme-ov-file#_restorestateonlanguagechange-boolean) on language change, but the current location was no longer within the `_startIds`, that location will be restored unless `_start._force: true`. This is because `isRouteSpecified` is `true` so the start location is [ignored](https://github.com/adaptlearning/adapt-contrib-core/blob/c7c74d45b3cf0751455658538a585d41db0bf340/js/startController.js#L57C5-L57C90).

This requires the author to understand that the `_force` value should be `true` whenever not restoring the state. Should this be something that is automatically applied to prevent the ability to return to a invalid location in these cases?

PR #707, opened 2025-10-17
Fix: Defer data-ready events

### LINKED ISSUE #698: `data.isReady` not reset on language change
When switching languages, `data.isReady` remains `true` despite the data being reloaded. Depending on the when the call is made, this can cause plugins directly calling `data.whenReady()` to resolve the promise early before models are ready, potentially causing errors.

joe-replin:
When switching language or reloading a course, plugins like trickle, page-level progress, and assessment sometimes run before the course models are fully built. This causes runtime errors (e.g., `findDescendantModels is not a function`) because plugins try to use incomplete model objects.

Originally reported here: https://github.com/adaptlearning/adapt-contrib-languagePicker/issues/120

### Fix
- Possibly fixes #698 & #705 
- Added `await wait.queue()` after `setupModel()` in `triggerDataLoaded()` so all models finish building before firing `app:dataLoaded`
- Removed duplicate event triggers in `onLanguageChange()`; now only triggered after models are ready

### Testing
You might not be able to replicate with a vanilla install of Adapt. I'm guessing it needs a wealth of content & features in it to replicate. Alas, this is how race conditions go. :(
1. Open a course with multiple languages
2. Progress to content using trickle, page-level progress, or assessment
3. Switch languages or reload the page mid-course
4. Check the browser console for errors
5. Confirm all plugins work as expected after switching or reloading

### Additional notes
This update makes sure plugins only run after all models are ready, preventing errors and improving reliability during language changes or reloads.

### DIFF
--- a/js/data.js
+++ b/js/data.js
@@ -88,7 +88,7 @@ class Data extends AdaptCollection {
     if (Adapt.get('_isStarted')) {
       Adapt.set('_isStarted', false);
     }
-    this.loadCourseData(language, previousLanguage);
+    await this.loadCourseData(language, previousLanguage);
   }
 
   async loadCourseData(newLanguage, previousLanguage) {
@@ -188,6 +188,7 @@ class Data extends AdaptCollection {
     try {
       // Setup the newly added models
       this.forEach(model => model.setupModel?.());
+      await wait.queue();
       Adapt.trigger('app:dataLoaded');
     } catch (e) {
       logging.error('Error during app:dataLoading trigger', e);


js/data.js
```diff
@@ -88,7 +88,7 @@ class Data extends AdaptCollection {
     if (Adapt.get('_isStarted')) {
       Adapt.set('_isStarted', false);
     }
-    this.loadCourseData(language, previousLanguage);
+    await this.loadCourseData(language, previousLanguage);
```
oliverfoster: This isn't necessary. 

`onLanguageChange` is triggered by a backbone event, which doesn't wait for promises to complete. So there is nothing waiting for `onLanguageChange` to complete and therefore no need to await the final call to `loadCourseData`. You can let this function-call hang in the breeze.

Equally, leaving this in isn't a problem. It just doesn't do anything.

cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

PR #815, opened 2026-01-07
Fix: Harden execution condition typo (fixes #814)

### LINKED ISSUE #814: Incorrect condition in core hardening
### Subject of the issue

This pr introduced code which should only run when not in development mode. But it does the opposite.

https://github.com/adaptlearning/adapt-contrib-core/pull/701

oliverfoster:
fixes #814 

Harden mistakenly had the opposite run condition.

### Fix
* Fixed the condition of execution by inverting !==

### Testing
Please also test and approve https://github.com/adaptlearning/adapt-contrib-spoor/pull/350
* `config.json:_fixes._harden = true`
* Must be built with `grunt build` or equivalent. 
* Check for `require('core/js/adapt')` in the console. It should not be accessible.

### DIFF
--- a/js/fixes/harden.js
+++ b/js/fixes/harden.js
@@ -8,7 +8,7 @@ import Adapt from 'core/js/adapt';
 Adapt.on('adapt:start', () => {
   const config = Adapt.config.get('_fixes');
   if (config?._harden !== true) return;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

PR #817, opened 2026-01-08
New: Add priority labels to templates and schemas (fix #816)

### LINKED ISSUE #816: Add priority labels to headings
Add an option to show priority labels (required/optional) in headings of the following content types: `_menuItem`, `_page`, `_article`, `_block`, `_component`.

Each content type can independently control whether labels are shown in its headings. For example, you could display "optional" labels on menu items but not on components.

Priority strings have already been added [here](https://github.com/adaptlearning/adapt-contrib-core/pull/521). This issue focuses specifically on:

- Adding the labels to heading areas
- Controlling when they are displayed for each content type

Note that displaying the labels for Box Menu items will be handled in another [PR](https://github.com/adaptlearning/adapt-contrib-boxMenu/pull/193).

See the original priority discussion [here](https://github.com/adaptlearning/adapt_framework/issues/3557).

### Related pull requests
The following PRs depend on this issue:

- Box menu priority label: https://github.com/adaptlearning/adapt-contrib-boxMenu/pull/193
- Vanilla label styling: https://github.com/adaptlearning/adapt-contrib-vanilla/pull/567

swashbuck:
Fix #816 

### New
* Adds priority labels to headings. Includes menu items, sub menus, content objects, articles, blocks, and component headings.

### Notes

- Can be configured to show optional labels, required labels, both, or none.
- Each content type can have different behavior for showing priority labels
- Priority label settings can be overridden on specific instances. For instance, you can disable labels globally for blocks but then show them for a specific block.
- Classes have been added so that the type of priority label can be styled differently (e.g. red labels for required and orange for optional)
- Priority labels only appear when a displayed title is present on the item.
- Read the new [wiki page](https://github.com/adaptlearning/adapt_framework/wiki/Priority-labels)

### Dependencies
* For styling: https://github.com/adaptlearning/adapt-contrib-vanilla/pull/567
* For Box Menu: https://github.com/adaptlearning/adapt-contrib-boxMenu/pull/193

### Screenshots
See https://github.com/adaptlearning/adapt-contrib-vanilla/pull/567

### Configuration
In _course.json_, add the following to `_globals` and enable/disable labels as desired.

The following example configuration will show optional labels only for menu item and page headings. In _contentObjects.json_, set `"_isOptional": true` to make content objects optional.

```
"_priorityLabels": {
  "_iconClassOptional": "",
  "_iconClassRequired": "icon-exclamation",
  "_menu": {
    "_showWhenOptional": true,
    "_showWhenRequired": false
  },
  "_menuItem": {
    "_showWhenOptional": true,
    "_showWhenRequired": false
  },
  "_page": {
    "_showWhenOptional": true,
    "_showWhenRequired": false
  },
  "_article": {
    "_showWhenOptional": false,
    "_showWhenRequired": false
  },
  "_block": {
    "_showWhenOptional": false,
    "_showWhenRequired": false
  },
  "_component": {
    "_showWhenOptional": false,
    "_showWhenRequired": false
  }
},
```

You can also override global priority label settings for individual elements. Add to the specific instance (ex. a block):

```
"_priorityLabels": {
    "_isOverride": true,
    "_showWhenRequired": true,
    "_showWhenOptional": false
},
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "adapt-contrib-core",
-  "version": "6.73.1",
+  "version": "6.74.0",
   "framework": ">=5.20.2",
   "description": "Adapt Framework kernel. Responsible for boot loading, data loading and core APIs.",
   "repository": {
--- a/templates/image.jsx
+++ b/templates/image.jsx
@@ -40,8 +40,7 @@ export default function Image(props) {
           props?._srcFocalPoint && `object-position-${props?._srcFocalPoint}`
         ])}
         src={src}
-        aria-label={props.alt ? a11y.normalize(props.alt) : null}
-        aria-hidden={!props.alt}
+        alt={a11y.normalize(props.alt)}
         loading='eager'
         aria-describedby={props.longdescription ? props.longDescriptionId : undefined}
         draggable={props.draggable ?? null}
--- a/templates/notifyPopup.hbs
+++ b/templates/notifyPopup.hbs
@@ -19,7 +19,7 @@
             {{#if _graphic._src}}
             {{#any (equals _imageAlignment 'left') (equals _imageAlignment 'top')}}
             <div class="notify__image-container{{#if _graphic.attribution}} has-attribution{{/if}}">
-              <img class="notify__image" src="{{_graphic._src}}" {{#if _graphic.alt}}aria-label="{{{_graphic.alt}}}"{{else}}aria-hidden="true"{{/if}}>
+              <img class="notify__image" src="{{_graphic._src}}" alt="{{{_graphic.alt}}}">
               {{#if _graphic.attribution}}
               <div class="notify__attribution">
                 <div class="notify__attribution-inner">
@@ -52,7 +52,7 @@
             {{~#if _graphic._src~}}
             {{~#any (equals _imageAlignment 'right') (equals _imageAlignment 'bottom')~}}
             <div class="notify__image-container{{#if _graphic.attribution}} has-attribution{{/if}}">
-              <img class="notify__image" src="{{_graphic._src}}" {{#if _graphic.alt}}aria-label="{{{_graphic.alt}}}"{{else}}aria-hidden="true"{{/if}}>
+              <img class="notify__image" src="{{_graphic._src}}" alt="{{{_graphic.alt}}}">
               {{#if _graphic.attribution}}
               <div class="notify__attribution">
                 <div class="notify__attribution-inner">


js/views/adaptView.js
```diff
@@ -419,6 +422,38 @@ class AdaptView extends Backbone.View {
...
+    const requiredLabel = _globals?._accessibility?._ariaLabels?.required;
+
+    const showWhenOptional = _priorityLabels._showWhenOptional && _isOptional && optionalLabel;
+    const showWhenRequired = _priorityLabels._showWhenRequired && !_isOptional && requiredLabel;
+
+    if (!showWhenOptional && !showWhenRequired) return;
+
+    // Set priority data on model
+    this.model.set({
+      priorityClass: _isOptional ? 'is-optional' : 'is-required',
```
oliverfoster:
```suggestion
      _priorityClass: _isOptional ? 'is-optional' : 'is-required',
```
Even though it isn't a config property from the json, the convention is still that non-translatable non-text is prefixed with `_`.

oliverfoster (changes requested):

js/views/adaptView.js
swashbuck: Updated.

js/views/adaptView.js
```diff
@@ -419,6 +422,38 @@ class AdaptView extends Backbone.View {
...
 
+  /**
+   * Calculate and set priority label data on the model based on global configuration.
+   * Handles all the logic internally: checks type, gets config, and sets model data.
+   */
+  setPriorityLabels() {
+    const type = this.constructor.type;
+
+    // Early exit if not a supported content type
+    if (!['menuItem', 'page', 'article', 'block', 'component'].includes(type)) {
```
joe-replin:
```suggestion
    const SUPPORTED_TYPES = ['menuItem', 'page', 'article', 'block', 'component'];
    const type = this.constructor.type;

    // Early exit if not a supported content type
    if (!SUPPORTED_TYPES.includes(type)) {
      return;
    }
```
Consider extracting supported types for maintainability.

joe-replin:
```suggestion
    const SUPPORTED_TYPES = ['menuItem', 'page', 'article', 'block', 'component'];
    const type = this.constructor.type;

    // Early exit if not a supported content type
    if (!SUPPORTED_TYPES.includes(type)) {
```
Consider extracting supported types for maintainability.

joe-replin (changes requested):

swashbuck:
Thanks, @kirsty-hames .

When a displayTitle isn't present:
1. Should we show the priority label at all? Currently, it's shown as long as there is a displayTitle, body text, or instruction text. So, an optional Graphic without any text will not show the label.
2. If we do show the label when there is no title, how should we announce the status to assistive technologies? Would they read the onscreen label or would we add a new AT-only span tag?

Screenshots showing 3 optional components: **Text**, **MCQ** and **Graphic**. Note that the label is _not_ shown for the optional **Graphic**.

kirsty-hames:
Hey @swashbuck, I've just tested the latest and can confirm the overrides at element level work as expected. 

One thing I did spot though is the component level priority labels don't apply to Media. I think this may be the only component still using the [_component.hbs_](https://github.com/adaptlearning/adapt-contrib-core/blob/master/templates/partials/component.hbs) instead of _header.jsx_ - should we update this too?

cahirodoherty-learningpool (approved):
Looks good 👍

PR #820, opened 2026-01-12
Fix: Logic for detecting removed elements (fixed #819)

### LINKED ISSUE #819: DOMElementModifications not triggering remove on remove+change
### Subject of the issue
If the element is removed and then changed, such that a src attribute on a video element is changed after removal, [the event name calculation function](https://github.com/adaptlearning/adapt-contrib-core/blob/a722aa19a55793f1502fd7a17339f3def23ca182/js/DOMElementModifications.js#L94-L107) will remove the `removed` event from the trigger list and return only the `changed` event.

oliverfoster:
fixes #819 

### Fix
* Enable `removed` events to trigger correctly on `removed + changed`


cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

PR #825, opened 2026-01-29
Update: use alt instead of aria-label for image alternative text (fixes #822)

### LINKED ISSUE #822: Should we use alt instead of aria-label for image alternative text?
### Subject of the issue
`alt` is a fundamental part of the HTML content model that serves multiple purposes beyond providing an accessible name.
`aria-label` is purely for the accessibility tree and used by assistive technologies only.

[Adapt images currently use the `aria-label` for alternative text](https://github.com/adaptlearning/adapt-contrib-core/blob/08676105d5f9521fddb525e7335ba61aed0ed5bb/templates/image.jsx#L43) however best practice advises using `alt` - always use native html over ARIA. I _**think**_ the initial reason `aria-label` was used in favour of `alt` in Adapt was due to inconsistencies in screen reader support and the handling of empty `alt` tags. As support has improved over the years I think it's worth reviewing and switch over to `alt` for images if this proves beneficial.

#### The `alt` attribute has multiple uses:
- **Screen readers**: Announced to assistive technology users (the primary reason why we provide alternative text in Adapt and the WAI  recommended implementation of alternative text for images)
- **Copy/paste**: When you copy an image and paste into a text editor (Word etc), many applications insert the alt text (useful when creating blended learning, accessible PDF versions of courses, or reviewing content)
- **Bookmarking**: Some browsers use alt text when you bookmark/save a linked image
- **Image search**: Search engines index alt text for image search results (less applicable for Adapt use cases perhaps)
- **Failed loads**: Browsers display alt text when images don't load

#### `aria-label` only affects the accessibility tree:

- Screen readers and other assistive technologies can access it
- It's invisible to all other contexts (copy/paste, bookmarks, etc.)
- It doesn't appear when images fail to load (loss of content)

Whilst both `alt` and `aria-label` provide an accessible name, `alt` provides a content alternative. Out of the multiple uses `alt` text has, I think having text display as an alternative for an image is perhaps the most beneficial for Adapt for the following use cases:

- Network and connection issues
- Broken link/file path, missing file
- User settings such as device low power mode or images disabled

#### Before proposing any amends, the following should be tested:

- [x] Screen reader compatibility across the different screen reader and browser combination 
- [x]    ^ Both populated `alt` and empty `alt=""`
- [x]    ^ Consistency in reading across screen reader and browser combinations 
- [x]  Compatibility with other ARIA - for example, [the image template currently supports `aria-describedby`](https://github.com/adaptlearning/adapt-contrib-core/blob/08676105d5f9521fddb525e7335ba61aed0ed5bb/templates/image.jsx#L46)

**References**: [MDN  alt](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#alt), [WAI images tutorial](https://www.w3.org/WAI/tutorials/images/)

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-core/issues/822

### Update
- replace image `aria-label` with `alt`
- remove image `aria-hidden` 

[Testing](https://github.com/adaptlearning/adapt-contrib-core/issues/822#issuecomment-3804968510) confirmed all major screen readers handle `alt` and `aria-label` equivalently and Adapt would benefit from the broader benefits of `alt`. Benefits include:
-  alternative text displayed when image fails to load 
- alternative text retained when copying images into Word or similar tools.

Testing also confirmed an empty alt tag (`alt=""`) is widely supported by screen readers / browsers to ignore an image without the need for `aria-hidden`.

### Related
- Boxmenu header image - https://github.com/adaptlearning/adapt-contrib-boxMenu/pull/231
- Hotgraphic main image - https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/357
- LanguagePicker image - https://github.com/adaptlearning/adapt-contrib-languagePicker/pull/127
- Tutor feedback image - https://github.com/adaptlearning/adapt-contrib-tutor/pull/123

### DIFF
--- a/templates/notifyPopup.hbs
+++ b/templates/notifyPopup.hbs
@@ -19,7 +19,7 @@
             {{#if _graphic._src}}
             {{#any (equals _imageAlignment 'left') (equals _imageAlignment 'top')}}
             <div class="notify__image-container{{#if _graphic.attribution}} has-attribution{{/if}}">
-              <img class="notify__image" src="{{_graphic._src}}" alt="{{_graphic.alt}}">
+              <img class="notify__image" src="{{_graphic._src}}" alt="{{{_graphic.alt}}}">
               {{#if _graphic.attribution}}
               <div class="notify__attribution">
                 <div class="notify__attribution-inner">
@@ -52,7 +52,7 @@
             {{~#if _graphic._src~}}
             {{~#any (equals _imageAlignment 'right') (equals _imageAlignment 'bottom')~}}
             <div class="notify__image-container{{#if _graphic.attribution}} has-attribution{{/if}}">
-              <img class="notify__image" src="{{_graphic._src}}" alt="{{_graphic.alt}}">
+              <img class="notify__image" src="{{_graphic._src}}" alt="{{{_graphic.alt}}}">
               {{#if _graphic.attribution}}
               <div class="notify__attribution">
                 <div class="notify__attribution-inner">


cahirodoherty-learningpool (approved):
👀

PR #834, opened 2026-02-26
Fix: Fix migration scripts issues (fix #833)

### LINKED ISSUE #833: Bug: Migration issue with _accessibility._isEnabled
### Subject of the issue
`stringToBoolean` migration helper overwrites existing boolean `true` values with `false` during v2.2.0 migration

### Your environment
* Framework v2.2.0 migration (`src/core/migrations/v2.js`)

### Steps to reproduce
1. Have a course where one or more boolean config fields (e.g. `config._accessibility._isEnabled`) are already stored as a proper boolean `true` (not the string `"true"`)
2. Run the v2.2.0 migration

### Expected behaviour
Fields that are already correct booleans should be preserved as-is.

### Actual behaviour
The `stringToBoolean` helper (line 310 of `v2.js`) only returns `true` when the input is the string `"true"`. When it receives a non-string value — including an already-correct boolean `true` — it falls through to return the `defaultValue`. For `_isEnabled`, the default is `false`, so any course with `_isEnabled: true` has it silently overwritten to `false` after migration.

### Suggested fix
Add a boolean passthrough at the top of `stringToBoolean`:

```js
function stringToBoolean(str, defaultValue) {
  if (_.isBoolean(str)) return str;
  if (!_.isString(str)) return defaultValue;
  return str.toLowerCase() === 'true';
}
```

swashbuck:
Fix #833 

### Fix
* Fix `stringToBoolean` issue in migration scripts to account for values that are already boolean.
* Fix `_isHiddenFromMenu` error message
* Fix `_navOrder` script

### DIFF
--- a/migrations/v2.js
+++ b/migrations/v2.js
@@ -308,6 +308,7 @@ describe('core - update to v2.2.0', async () => {
 
   // Convert strings to boolean or return default value
   function stringToBoolean(str, defaultValue) {
+    if (_.isBoolean(str)) return str;
     if (!_.isString(str)) return defaultValue;
     return str.toLowerCase() === 'true';
   }
--- a/migrations/v3.js
+++ b/migrations/v3.js
@@ -310,7 +310,7 @@ describe('core - update to v3.1.0', async () => {
 
   checkContent('core - check contentObjects._isHiddenFromMenu', async (content) => {
     const isValid = contentObjects.every(contentObject => !_.has(contentObject, '_isHiddenFromMenu'));
-    if (!isValid) throw new Error('core - contentObjects._isHiddenFromMenu not added');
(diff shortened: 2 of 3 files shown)


oliverfoster (approved):

swashbuck:
@oliverfoster Added one more fix for `_navOrder`. That should be it for now.

cahirodoherty-learningpool (approved):

mohosy (commented):
Quick review pass on 'Fix: Fix migration scripts issues (fix #833)': changedFiles=3, additions=6, deletions=5.\n\nPlease confirm regression coverage for the touched paths (especially edge/error handling), and consider adding a brief note for any behavior-visible changes.

PR #847, opened 2026-04-13
Fix: Maximum stack exceeded error

### LINKED ISSUE #831: Recursive loop error when focusing elements
It is possible for the following error to occur when focus is changed when interacting with elements:
```JavaScript
a11y.js:800 Uncaught RangeError: Maximum call stack size exceeded.
```
This seems to occur due to the changes made in https://github.com/adaptlearning/adapt-contrib-core/pull/697. When [`_refocusCurrentActiveElement`](https://github.com/adaptlearning/adapt-contrib-core/blob/01a99626eb233a470f499fe3633a95ffd460d3e9/js/a11y/browserFocus.js#L62) is called following a blur event, a recursive loop can occur as the focus switches and `_onBlur` is subsequently called again with the `event.relatedTarget` set to `null`, bypassing the [guard](https://github.com/adaptlearning/adapt-contrib-core/blob/01a99626eb233a470f499fe3633a95ffd460d3e9/js/a11y/browserFocus.js#L50-L53). This doesn't occur with every focus switch.

This can be resolved by adding the defer option when refocussing the current active element, giving the task a chance to settle without firing subsequent events causing the recursive loop:
``` JavaScript
_refocusCurrentActiveElement() {
  const element = this.a11y.currentActiveElement;
  if (!element) return;
  // refocus on the existing active element to stop jaws from scrolling
  this.a11y.focus(element, { preventScroll: true, defer: true });
}
```

oliverfoster:
fixes #831 

### Cause

1. When forcing focus onto an element with `a11y.focus`, the attributes `[tabindex=0, data-a11y-force-focus=true]` are added to the subject element, allowing elements which are not natively focus to receive focus. On blur when `[tabindex=0]` is removed, if the element is not natively focusable, such as a `div[role=heading]`, then the focus is lost, causing a refocus.
2. When clicking a label element in an mcq, refocus was being triggered as the label is not natively focusable and nothing in its ancestor stack is focusable.

### Replication

When using trickle to move to the next question, where trickle focuses the next heading and a user subsequently clicks an mcq label, a refocus loop is created. 

### Fix

* Only removing the `[tabindex]` from the `div[role=heading]` after focus is reassigned where the focus is on a different element.
* Allow label clicks to be considered as valid, non-focusable clicks.

### DIFF
--- a/js/a11y/browserFocus.js
+++ b/js/a11y/browserFocus.js
@@ -62,18 +62,21 @@ export default class BrowserFocus extends Backbone.Controller {
       return;
     }
     const $element = $(event.target);
-    if ($element.is('[data-a11y-force-focus]')) {
+    const finish = () => {
+      const hasFocus = ($element[0] === this.a11y.currentActiveElement);
+      if (hasFocus) return;
+      if (!$element.is('[data-a11y-force-focus]')) return;
       $element.removeAttr('tabindex data-a11y-force-focus');
-    }
+    };
     // From here, only check source elements
     if (event.target !== event.currentTarget) {
-      return;
+      return finish();
     }
     // Do not auto next if the focus isn't returning to the body or html element
     // or if we're not losing focus
     const isNotBodyHTMLOrLostFocus = (!$(event.relatedTarget).is('body, html') && event.relatedTarget !== null);
     if (isNotBodyHTMLOrLostFocus) {
-      return;
+      return finish();
     }
     // Check if element is losing focus
     // due to the addition of a disabled class, display none, visibility hidden,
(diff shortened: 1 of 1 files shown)


lemmyadams (approved):

danielghost (approved):
Issue resolved and has been tested by QA in actual projects

swashbuck (approved):
👍 

Tested locally against this branch (`issue/831`) on framework v5.56.2, core built from `423ed27`, with a trickle + question repro page (`_accessibility._isEnabled: true`).

### Repro steps

1. Trickle-step from an MCQ block to the next block, which has a `displayTitle` (so trickle's `a11y.focusFirst` force-focuses the block's `div[role=heading]`).
2. Click a GMCQ option label in the newly revealed block.

### Results

- **Before** (master @ `2c6ef07`): the label click triggers `Uncaught RangeError: Maximum call stack size exceeded` (the refocus loop described in #831).
- **After** (this branch): no error. Focus moves cleanly to the clicked control; repeated MCQ/GMCQ label clicks with both headings force-focused stay stable.

Verified the fix is the differentiator: identical course content and steps, only the core bundle swapped between `master` and `issue/831`.

Posted via collaboration with Claude Code

PR #866, opened 2026-05-19
Update: Add new IcoMoon config, readme, and font files (fix #713)

### LINKED ISSUE #713: Icomoon - New app and core icon font
### Subject of the issue
Icomoon has a [new app](https://icomoon.io/new-app) in _beta_. We should look at the Core icon fonts and determine if the `selection.json` [file](https://github.com/adaptlearning/adapt-contrib-core/blob/master/fonts/selection.json) should be updated or if there are any additional Icomoon files that we should keep here.

https://github.com/adaptlearning/adapt-contrib-core/blob/master/fonts/

swashbuck:
Fix #713 

### Update
* Add new configuration files from the **new** [IcoMoon app](https://icomoon.io/new-app). Keep _selection.json_ for backwards compatibility.
* Add _README.md_ to help with updating the icon font.
* Update icon font files (WOFF, WOFF2). Exported from new IcoMoon app.

### Considerations
IcoMoon app is still in beta, so some things may change. However, the exported font files should still work here. The _selection.json_ config will work in the old app as a fallback.

### Firefox issues with font

The new IcoMoon app exports CFF2 outlines instead of TrueType. Its CFF2 encoder has a bug: for two glyphs with complex multi-subpath geometry (wechat, youku), it emits `rmoveto` with extra stack operands. The CFF2 spec requires exactly 2 operands; Firefox's OpenType Sanitizer enforces this strictly and rejects the font. Chrome and Safari are lenient and were unaffected.

These are the SVGs with invalid paths:
wechat
youku

#### Fix for Firefox

Fixed the two affected glyphs in Illustrator:

1. Select all
2. Pathfinder > Unite
3. Manually deleted the stray path layers that now appeared.
4. Object > Path > Clean Up to be safe

Illustrator

Once fixed, I replaced them in the IcoMoon project before exporting. The exported fonts were validated with `ots-sanitize` (the same OpenType Sanitizer Firefox uses). All 246 glyphs now pass.

#### Note for future font updates

Run `ots-sanitize vanilla.woff2` after every IcoMoon export before committing. If any glyph trips the encoder bug, it will report `ERROR: CFF2: Failed validating CharStrings INDEX`. A bug report has been filed with IcoMoon.

### Testing

- Chrome, Safari, Firefox: all icons render correctly
- `ots-sanitize vanilla.woff2` reports "File sanitized successfully"
- Codepoints unchanged (`\e900`–`\e9f1`) — no changes to `icons.less`

#### Verifying with ots-sanitize

Install once:

Mac / Linux:
```bash
pip3 install --user opentype-sanitizer
```

Windows (Command Prompt or PowerShell):
```
pip install --user opentype-sanitizer
```

Run against the font files (from the repo root):

Mac / Linux:
```bash
python3 -m ots src/core/fonts/vanilla.woff2
python3 -m ots src/core/fonts/vanilla.woff
```

Windows:
```
python -m ots src\core\fonts\vanilla.woff2
python -m ots src\core\fonts\vanilla.woff
```

Expected output for both: `File sanitized successfully!`

If a glyph is invalid, you'll see:

```bash
ERROR: CFF2: Failed validating CharStrings INDEX
ERROR: CFF2: Failed to parse table
Failed to sanitize file!
```

There is nothing off-the-shelf that I could find that would identify _which_ glyphs are invalid. I do have a local Python script that will do this, so ping me if you need it.

_Update 2026-05-20:_ IcoMoon says that the issue has been resolved.


kirsty-hames (approved):
Works as expected thanks @swashbuck 👍 

Testing on macOS Safari, FF and Chrome with both a new Adapt setup and updating an existing project with custom font icons in the theme.

Sanitize prompts successful too.

PR #880, opened 2026-08-19
Update: Drop WOFF2 icon font (fixes #877)

### LINKED ISSUE #877: IcoMoon drops free support for WOFF2
## Issue
Adapt core currently uses [IcoMoon](https://icomoon.io/pricing) for creating the WOFF and WOFF2 icon fonts. Unfortunately, the free option no longer included the _WOFF2_ format.

>An active paid plan is required for WOFF2 generation. See https://icomoon.io/pricing

Suggest that we just drop the WOFF2 font altogether. The main difference is file size, but WOFF is still a reasonably small size.

* vanilla.woff - 31KB
* vanilla.woff2 - 18.5KB

https://github.com/adaptlearning/adapt-contrib-core/tree/master/fonts

swashbuck:
Fixes #877

### Update
- Removed the WOFF2 icon font. IcoMoon's free plan no longer generates WOFF2, so the file could not be kept in step with future icon-set changes. The `@font-face` rule now loads WOFF only, which every supported browser handles.
- Updated the fonts _README.md_ so the regeneration steps no longer ask for a WOFF2 export, and noted why one should not be added back.

### Performance Impact
The icon font download grows from 18.5KB to 31KB. It is a single request, cached after first load, so the effect is negligible.

### Testing
1. Build a course and load it in a browser.
2. Confirm all icons render as before (navigation bar, drawer, buttons, accordion/narrative controls).
3. Check the network panel: only _vanilla.woff_ is requested, and there is no 404 for _vanilla.woff2_.

Posted via collaboration with Claude Code


ethan-lp (approved):

kirsty-hames (approved):
Works as expected thanks @swashbuck 👍 Testing across Safari, Firefox and Chrome on macOS

